PR #199, opened 2020-10-02
updateFile method added to projects

jk05:

Adds in an updateFile method to Projects, which wraps around addFile and overwrites. On Common and Web packages

currently cannot overwrite files with `addFile`

### What is the new behavior?
Allows files to be updated by overwriting existing files

Created a card to add this to CLI package later (https://trello.com/c/EekG2aPk)

### DIFF
--- a/packages/common/src/entities/projects/projects.abstract.ts
+++ b/packages/common/src/entities/projects/projects.abstract.ts
@@ -53,7 +53,20 @@ export abstract class ProjectsAbstract<Env extends EnvironmentAbstract> {
      * @param   source
      * @param   destination
      */
-    abstract addFile(projectId: string, source: string, destination?: string): Promise<IRelateFile>;
+    abstract addFile(
+        projectId: string,
+        source: string,
+        destination?: string,
+        overwrite?: boolean,
+    ): Promise<IRelateFile>;
+
+    /**
+     * Update a file (overwrite) in a project
+     * @param   projectId
+     * @param   source
+     * @param   destination
+     */
+    abstract updateFile(projectId: string, source: string, destination?: string): Promise<IRelateFile>;
 
     /**
      * Adds file (write) to project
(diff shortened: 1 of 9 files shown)


packages/common/src/entities/projects/projects.abstract.ts
```diff
@@ -53,7 +53,20 @@ export abstract class ProjectsAbstract<Env extends EnvironmentAbstract> {
      * @param   source
      * @param   destination
      */
-    abstract addFile(projectId: string, source: string, destination?: string): Promise<IRelateFile>;
+    abstract addFile(
+        projectId: string,
+        source: string,
+        destination?: string,
+        overwrite?: boolean,
+    ): Promise<IRelateFile>;
```
huboneo, line 61: Looking at it now I cant help but wonder if it would be better not have the `overwrite` flag and just copy over the logic needed to updateFile. @nglgzz thoughts?

huboneo (approved):
LGTM, but please run `generate:docs` before committing

One minor question, not a blocker

packages/common/src/entities/projects/projects.abstract.ts
nglgzz: I was actually going to ask the opposite. Wouldn't it make sense to keep the overwrite flag and remove the duplicate `updateFile` method? Why would we want them separate?

huboneo: Good point, either way works for me but I feel they are duplicate logic at the moment

huboneo: In that case do we need to guard for if the file does not exist, @jas maybe just use `overwrite` when file already exists?

jk05: We originally spoke about the semantics of having add and update methods as separate. But I am happy to go either way,

jk05: so shall we remove `updateFile` then and just overwrite if a file exists?

jk05: Or pass in overwrite flag explicitly as @nglgzz suggested? Ill wait on your thoughts.

huboneo: lets go with @nglgzz suggestion

huboneo (approved):

PR #201, opened 2020-10-06
Use correct host URL when verifying API tokens

nglgzz:

- Fix the API token verification
- Pass the correct parameters to `dbms.upgrade` in the CLI module
- Use platform-independent paths when upgrading DBMSs
- Change the `--confirm` flag to boolean type on `dbms:link`

When verifying API tokens we use `req.hostname` as part of the token salt, which contains the origin URL in the format `<hostname>`, while when generating the tokens we use `<hostname>:<port>`. This means that currently all requests are rejected as unauthorized, even when they should be valid.

### What is the new behavior?
Tokens are now verified properly.

The `--confirm` flag on `dbms:link` is now a boolean instead of a string, so instead of calling the command as:
```
relate dbms:link <path> <name> --confirm=confirm
```
will have to be called as:
```
relate dbms:link <path> <name> --confirm
```

### DIFF
--- a/packages/common/src/utils/dbmss/dbms-upgrade-config.ts
+++ b/packages/common/src/utils/dbmss/dbms-upgrade-config.ts
@@ -1,4 +1,5 @@
 import fse from 'fs-extra';
+import path from 'path';
 import semver from 'semver';
 
 import {IDbmsInfo} from '../../models';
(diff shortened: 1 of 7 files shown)


packages/common/src/utils/dbmss/dbms-upgrade-config.ts
```diff
@@ -13,28 +14,39 @@ export async function dbmsUpgradeConfigs(
...
+    const copyDBMSPath = (sourceRootPath: string, destinationRootPath: string, ...paths: string[]) =>
+        fse.copy(path.join(sourceRootPath, ...paths), path.join(destinationRootPath, ...paths));
+
+    await copyDBMSPath(dbms.rootPath, upgradedDbms.rootPath, 'data');
+    await copyDBMSPath(dbms.rootPath, upgradedDbms.rootPath, 'logs');
+    await copyDBMSPath(dbms.rootPath, upgradedDbms.rootPath, 'conf');
+    await copyDBMSPath(dbms.rootPath, upgradedDbms.rootPath, 'plugins');
 
-    const certsExists = await fse.pathExists(`${dbms.rootPath!}/certificates`);
+    const certsExists = await fse.pathExists(path.join(dbms.rootPath, 'certificates'));
```
jk05, line 33: good spot. We should do a final check through to make sure we dont have any more cases where `path` should be used

jk05 (approved):

PR #203, opened 2020-10-08
enable cors and req.origin string null check

jk05:

Enable CORS on web package and do a origin header string check. Similar change being added to bootstrapping of Relate module in Desktop

If origin is string 'null', currently being used as such. CORS not enabled.

### What is the new behavior?
CORS will be enabled and origin header of string 'null' will be handled correctly

### DIFF
--- a/packages/web/src/auth/services/auth.service.ts
+++ b/packages/web/src/auth/services/auth.service.ts
@@ -49,7 +49,8 @@ export class AuthService {
             }
 
             // Use the client URL otherwise fallback to the Relate server URL.
-            const requestUrl = req.get('origin') || environment.httpOrigin;
+            const requestUrl =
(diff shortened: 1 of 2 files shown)


huboneo:
Is there a reason why we dont just add this to the web module? also why not use `cors` (and probably `helmet`) npm packages for this?

nglgzz:
Looking at the NestJS docs it seems like they're using `cors` under the hood here, and the `enableCors` method is available only in the Nest application object. 

https://docs.nestjs.com/techniques/security#cors

huboneo:
@nglgzz I see, in that case LGTM

huboneo (approved):
ᕕ( ᐛ )ᕗ

jk05:
Will let @nglgzz sort out the new release pipeline before merging this

PR #204, opened 2020-10-09
Bugfix config reset upgrade

huboneo:

This PR addresses a bug where DBMS config values would be reset to default during upgrade. This PR also adds a download progress bar when upgrading from CLI.

Custom DBMS config values are not preserved after upgrade

### What is the new behavior?
Custom DBMS config values are preserved after upgrade

No

### Other information:
Screenshot 2020-10-09 at 09 00 45


packages/common/src/entities/dbmss/dbmss.local.ts
```diff
@@ -205,9 +205,14 @@ export class LocalDbmss extends DbmssAbstract<LocalEnvironment> {
 
         try {
             const upgradedDbms = await this.install(upgradeTmpName, version, dbms.edition!, '', noCache);
-            const upgradedConfig = await this.getDbmsConfig(upgradedDbms.id);
+            const upgradedConfigFileName = path.join(
+                this.environment.getEntityRootPath(ENTITY_TYPES.DBMS, dbmsId),
+                DBMS_MANIFEST_FILE,
```
nglgzz, line 210: Why are we getting the manifest instead of the config now?

huboneo: because I'm obviously not awake. good catch, I'll fix!

packages/cli/src/modules/dbms/upgrade.module.ts
```diff
@@ -20,6 +21,21 @@ export class UpgradeModule implements OnApplicationBootstrap {
         @Inject(SystemProvider) protected readonly systemProvider: SystemProvider,
     ) {}
 
+    registerHookListeners() {
```
jk05, line 24: 🥇

packages/common/src/utils/dbmss/dbms-upgrade-config.ts
```diff
@@ -23,7 +23,7 @@ export async function dbmsUpgradeConfigs(
     }
 
     const copyDBMSPath = (sourceRootPath: string, destinationRootPath: string, ...paths: string[]) =>
-        fse.copy(path.join(sourceRootPath, ...paths), path.join(destinationRootPath, ...paths));
+        fse.copy(path.join(sourceRootPath, ...paths), path.join(destinationRootPath, ...paths), {overwrite: true});
```
jk05, line 26: what's the reason for the `overwrite` here? For potential failed upgrades?

huboneo: We upgrade by installing a new DBMS that will already have most of these files with default values

jk05: right of course 👍

jk05: so this wouldnt actually have worked before then...

jk05 (approved):
lgtm. @nglgzz may know some more specifics.

packages/common/src/utils/dbmss/dbms-upgrade-config.ts
huboneo: Now that I think about it this could be a dangerous change. For example our own security plugin would be overwritten. @nglgzz thoughts?

nglgzz: Good point. I think overwriting makes sense, as we'd be resetting the DBMS to its default value otherwise. 

I'm guessing we only want to copy the contents of those directories though, or an upgrade from 3.5 to 4.0 would remove the security plugin.

nglgzz: @huboneo will you add this change in this PR or in a separate one?

huboneo: Updated the PR to preserve JWT settings but update the plugin. According to FSE docs `copy` only copies content of directories https://github.com/jprichardson/node-fs-extra/blob/master/docs/copy.md#copysrc-dest-options-callback

nglgzz (approved):
Looks good to me :sparkles:

nglgzz (approved):

nglgzz:
[Upgrading Neo4j resets the config file](https://trello.com/c/BFpfN8Gg/1477-upgrading-neo4j-resets-the-config-file)

PR #209, opened 2020-10-13
build(cli) oclif packaging for standalone distribution 

akollegger:

Build improvement for @relate/cli to create standalone distributions which include a nodejs runtime.

@relate/cli is an npm package which requires nodejs.

### What is the new behavior?

Nodejs is now packaged along with the cli.

No changes expected to the functions of the cli. However, there is now the possibility that more than one @relate/cli is installed on the system (packaged with Desktop; installed with npm; installed as standalone cli).


huboneo:
Just realised that we call `npm install` when adding extensions, so we need to make sure NPM is available in PATH

akollegger:
@huboneo WDYT about switching to a library instead, like https://github.com/npm/libnpm ?

huboneo:
In general I'm against using anything non-standard. How can we guarantee that said library keeps up with NPM? Is there an official recommendation from oclif?

huboneo:
But libnpm does look like a good candidate should we not be able to use npm directly

PR #211, opened 2020-10-19
Fix DBMS upgrade not migrating data and not setting correct SSL configs

nglgzz:

Bug fixes

- When upgrading from 3.5.x to 4.x, the SSL certificates are copied in a different location. Due to a bad condition, the config is not updated to reflect that, and authentication always gets disabled.
- During any upgrade, where we try to migrate the data we stop the DBMS too fast after it's started, so the data is never migrated.

### What is the new behavior?
- When upgrading from 3.5.x to 4.x SSL certificates are copied as before, the config is updated accordingly, and authentication is disabled only when needed.
- When the `migrate` option is set to `true` data is migrated correctly.
- `IDbmsInfo` now contains a new field called `serverStatus`, which will be `DBMS_SERVER_STATUS.UNKNOWN` unless the `onlineCheck` option is passed to `dbmss.info`.

Not a breaking change but `IDmsInfo` now contains an extra field called `serverStatus`.

### Other information:
I was initially planning to add a new value to `DBMS_STATUS` but that change wasn't backward compatible and it would have broken several CLI commands and maybe some functions in Desktop. Besides that, I thought it would be better to separate the status of the DBMS process from the one of the DBMS server, as in most cases we don't need both, and the online check can be quite slow when the DBMS is starting (even without retries).


packages/common/src/utils/dbmss/is-dbms-online.ts
```diff
@@ -0,0 +1,62 @@
...
+
+    return reachable.toArray().every(Boolean);
+};
+
+const sleep = (ms: number): Promise<void> =>
+    new Promise((resolve) => {
+        setTimeout(resolve, ms);
+    });
+
+export const waitForDbmsToBeOnline = async (dbms: IDbms): Promise<void> => {
```
nglgzz, line 43: Not entirely satisfied with this name, so suggestions are welcome.

nglgzz:
[Starting a DBMS upgraded from 3.5 to 4.0 fails with "Starting Neo4j failed: HTTPS set to enabled, but no SSL policy provided"](https://trello.com/c/n33h2q0O/1480-starting-a-dbms-upgraded-from-35-to-40-fails-with-starting-neo4j-failed-https-set-to-enabled-but-no-ssl-policy-provided)

packages/cli/src/modules/dbms/info.module.ts
```diff
@@ -25,15 +25,17 @@ export class InfoModule implements OnApplicationBootstrap {
...
                 .mapEach((dbms) => {
                     return {
                         id: dbms.id,
                         name: dbms.name,
                         version: dbms.version,
                         edition: dbms.edition,
-                        status: dbms.status.trim(),
+                        status: dbms.status,
+                        // eslint-disable-next-line camelcase, @typescript-eslint/camelcase
+                        server_status: dbms.serverStatus,
```
huboneo, line 38: plz no underscores in JS :)

packages/common/src/utils/dbmss/dbms-upgrade-config.ts
```diff
@@ -51,21 +51,19 @@ export async function dbmsUpgradeConfigs(
...
-            path.join(upgradedDbms.rootPath, NEO4J_CERT_DIR, 'neo4j.cert'),
-            path.join(upgradedDbms.rootPath, NEO4J_CERT_DIR, 'https', 'neo4j.cert'),
-        );
-        await fse.copy(
-            path.join(upgradedDbms.rootPath, NEO4J_CERT_DIR, 'neo4j.key'),
-            path.join(upgradedDbms.rootPath, NEO4J_CERT_DIR, 'https', 'neo4j.key'),
-        );
-    }
-
     if (semver.lt(dbms.version!, NEO4J_VERSION_4) && semver.gte(upgradedDbms.version!, NEO4J_VERSION_4)) {
```
huboneo, line 54: are we sure this is only for >= 4.X?

packages/common/src/utils/dbmss/is-dbms-online.ts
```diff
@@ -0,0 +1,62 @@
...
+        [NEO4J_CONFIG_KEYS.HTTPS_CONNECTOR, DEFAULT_NEO4J_HTTPS_PORT],
+    ]);
+
+    const reachable = await connectors
+        .filter(([conn]) => conf.get(conn + NEO4J_CONFIG_KEYS.ENABLED) === 'true')
+        .mapEach(([conn, defaultPort]) => {
+            const address = conf.get(conn + NEO4J_CONFIG_KEYS.LISTEN_ADDRESS);
+
+            const protocol = conn.includes('https') ? 'https' : 'http';
+            const [, port] = (address || defaultPort).split(':');
```
huboneo, line 28: I understand this right, that this will pull the port if set in config?

packages/common/src/utils/dbmss/is-dbms-online.ts
```diff
@@ -0,0 +1,62 @@
...
+    let currentAttempt = 0;
+
+    /* eslint-disable no-await-in-loop */
+    while (currentAttempt < maxAttempts) {
+        const isOnline = await isDbmsOnline(dbms);
+        if (isOnline) {
+            return;
+        }
+        await sleep(delay);
+        currentAttempt += 1;
```
huboneo, line 55: IMO recursions are easier to read, but I leave it up to you

huboneo (approved):
LGTM overall, just some comments

packages/common/src/utils/dbmss/is-dbms-online.ts
nglgzz: Yep, otherwise it will use the defaults.

packages/cli/src/modules/dbms/info.module.ts
nglgzz: Yeah I know, it's just that without the underscore it will show as one word in the CLI table instead of two separate words.

packages/common/src/utils/dbmss/dbms-upgrade-config.ts
nglgzz: I'm not sure, but that's how Desktop was doing it. Either way, we're copying the whole certificates folder if it exists (even below 4.x), so I think here it's just a matter of moving those two files to a different path.

PR #218, opened 2020-10-24
Feature public extensions

huboneo:
Screenshot 2020-10-26 at 15 35 56

This PR removes the use of a private JFrog NPM registry in favor of the public NPM registry. A relate extension is now defined as any NPM package with the `neo4j-relate-extension` keyword. Extensions published under the `@relate/` NPM namespace are marked as "official", and all signed extensions are verified against the [Code Signing CRL Server](https://github.com/neo4j-devtools/code-signing-crl-server).

This PR also adds the ability to install extensions from a URL.

Only extensions published on the private registry can be installed

### What is the new behavior?
Any extensions published on the public registry can be installed

Yes, old extensions can no longer be used

### Other information:
This PR uses Neo4j Desktop CA certificate for now, we need to issue a new intermediate certificate for relate (i'd assume). Further we should review how extension manifests are shaped and used, making sure we're not painting ourselves into a corner here


packages/common/src/entities/extensions/extensions.constants.ts
```diff
@@ -0,0 +1,73 @@
+// @todo: Create cert for relate and decide where it should live
```
huboneo, line 1: This needs to be replaced with a new chain for relate

packages/common/src/entities/extensions/extensions.local.ts
```diff
@@ -162,17 +171,15 @@ export class LocalExtensions extends ExtensionsAbstract<LocalEnvironment> {
...
         const execute = promisify(exec);
 
-        try {
+        if (extension.type !== EXTENSION_TYPES.STATIC) {
             await emitHookEvent(
                 HOOK_EVENTS.RELATE_EXTENSION_DEPENDENCIES_INSTALL_START,
                 `installing dependencies for ${extension.name}`,
             );
-            const output = await execute('npm install --production', {
+            const output = await execute('npm ci --production', {
```
huboneo, line 179: we need to use CI to enable signed extensions to install dependencies, which means they have to be signed with their prod deps

packages/common/src/utils/extensions/extension-versions.ts
```diff
@@ -61,71 +49,67 @@ export async function discoverExtension(extensionRootDir: string): Promise<IExte
     }
 
     const extensionManifest = path.join(extensionRootDir, EXTENSION_MANIFEST_FILE);
-    const extensionManifestLegacy = path.join(extensionRootDir, EXTENSION_MANIFEST_FILE_LEGACY);
-    const hasManifest = await fse.pathExists(extensionManifest);
-    const hasManifestLegacy = await fse.pathExists(extensionManifest);
+    const hasManifestFile = await fse.pathExists(extensionManifest);
     const extensionPackageJson = path.join(extensionRootDir, PACKAGE_JSON);
     const hasPackageJson = await fse.pathExists(extensionPackageJson);
```
huboneo, line 55: Changes here are unrelated to the PR topic, just wanted to clean up how we detect extension meta

packages/common/src/utils/extensions/get-installed-extensions.ts
```diff
@@ -58,15 +58,15 @@ function mapContentsToExtension(fullPath: string): IInstalledExtension {
...
+            name: _.replace(manifest.name, RELATE_NPM_PREFIX, ''),
             root: fullPath,
         });
     }
 
     if (hasPackageJson) {
         return new ExtensionModel({
             ...contents,
-            name: _.replace(contents.name, EXTENSION_NPM_PREFIX, ''),
+            name: _.replace(contents.name, RELATE_NPM_PREFIX, ''),
```
huboneo, line 69: not sure we should keep these replaces, I believe we did it for legibility, WDYT @nglgzz

nglgzz: It's probably best to remove it since it could cause ambiguity with packages published outside the `@relate` namespace.

huboneo:
My one concern here is that we check the signature of an extension every time it is discovered (i.e. listed). This enables us to catch revocations on the fly, but slows down list operations due to network latency. WDYT @oskarhane ?

packages/common/src/entities/extensions/extensions.local.ts
```diff
@@ -82,38 +81,46 @@ export class LocalExtensions extends ExtensionsAbstract<LocalEnvironment> {
...
 
-        const {extensionsCache} = this.environment.dirPaths;
+        const {extensionsCache, tmp} = this.environment.dirPaths;
+        const isFilePath = (await fse.pathExists(version)) && (await fse.stat(version)).isFile();
 
-        // version as a file path.
-        if ((await fse.pathExists(version)) && (await fse.stat(version)).isFile()) {
-            // extract extension to cache dir first
+        // version as a file path or URL.
+        if (isFilePath || isValidUrl(version)) {
```
jk05, line 88: nice

packages/common/src/entities/extensions/extensions.local.ts
```diff
@@ -82,38 +81,46 @@ export class LocalExtensions extends ExtensionsAbstract<LocalEnvironment> {
...
+                await emitHookEvent(HOOK_EVENTS.RELATE_EXTENSION_DOWNLOAD_STOP, null);
+            }
+
+            // extract extension to tmp first
+            const tmpExtractPath = path.join(tmp, uuidv4());
             const {name: extensionName, dist, version: extensionVersion} = await extractExtension(
-                version,
-                extensionsCache,
+                toExtract,
+                tmpExtractPath,
```
jk05, line 105: Would we clean up this `tmpExtractPath` or did we say we would leave this to the OS to handle it for tmp?

packages/common/src/entities/extensions/extensions.local.ts
```diff
@@ -82,38 +81,46 @@ export class LocalExtensions extends ExtensionsAbstract<LocalEnvironment> {
...
+            await fse.move(dist, cacheDir, {
                 overwrite: true,
             });
 
             try {
-                const discovered = await discoverExtension(destination);
+                const discovered = await discoverExtension(cacheDir);
 
-                return this.installRelateExtension(discovered, discovered.dist);
+                return this.installRelateExtension(discovered, discovered.dist).finally(() => fse.remove(cacheDir));
```
jk05, line 117: guess this prevents building up cached extensions over time which is good

packages/common/src/entities/extensions/extensions.constants.ts
huboneo: Alternatively not have it comitted at all, rn relate has no hardcoded secrets

packages/common/src/utils/extensions/extension-versions.ts
```diff
@@ -135,62 +119,25 @@ export interface IExtensionVersion {
...
-                            name: `${extName}`,
-                            origin: EXTENSION_ORIGIN.ONLINE,
-                            version: `${version}`,
-                        };
-                    });
-                });
+        .mapEach(({name, origin, version}) => {
+            return {
+                name,
+                origin: origin || EXTENSION_ORIGIN.ONLINE,
```
nglgzz, line 138: Wouldn't this be always `EXTENSION_ORIGIN.ONLINE`? And in what cases do we have a value coming from NPM that matches our enum?

packages/common/src/models/extension.model.ts
```diff
@@ -27,3 +27,56 @@ export class ExtensionModel extends ModelAbstract<IInstalledExtension> implement
     @IsString()
     root!: string;
 }
+
+export class InstalledExtensionModel extends ModelAbstract<IInstalledExtension> implements IInstalledExtension {
+    @IsString()
+    name!: string;
```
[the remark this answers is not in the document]
huboneo: Tells typescript to skip null check

huboneo:
@nglgzz thank you for your feedback, I have updated the PR

packages/common/src/utils/extensions/extension-versions.ts
huboneo: No, this is mapping over the concatenated list of cached and online

packages/cli/docs/extension.md
```diff
@@ -17,7 +17,7 @@ USAGE
   $ relate extension:install [NAME]
 
 OPTIONS
-  -V, --version=version          Version to install
+  -V, --version=version          Version to install (semver), or path to tarball
```
oskarhane, line 20: 👍

packages/cli/src/modules/extension/install.module.ts
```diff
@@ -80,6 +83,11 @@ export class InstallModule implements OnApplicationBootstrap {
             const maybeWithName = name
                 ? _.filter([...choices.values()], (v) => v.name === name)
                 : [...choices.values()];
+
+            if (!arrayHasItems(maybeWithName)) {
+                throw new InvalidArgumentError(`Could not find installable extension with name: ${name}`);
+            }
```
oskarhane, line 89: Great!

packages/common/src/constants.ts
```diff
@@ -184,3 +183,11 @@ export enum FILTER_CONNECTORS {
...
 export const HEALTH_BASE_ENDPOINT = '/health';
 export const STATIC_APP_BASE_ENDPOINT = '/static';
+
+export enum EXTENSION_VERIFICATION_STATUS {
+    UNKNOWN = 'UNKNOWN',
+    UNSIGNED = 'UNSIGNED',
+    TRUSTED = 'TRUSTED',
+    UNTRUSTED = 'UNTRUSTED',
+    REVOKED = 'REVOKED',
+}
```
oskarhane, line 193: Awesome that you added verification to packages 🎍

oskarhane (approved):
I ran into some issues that I pointed out on the side and you fixed, thanks!

First, this is great, this is an important step.

I do think we will run into issues with conflicts with app-ids quite fast though, so we should think about how to solve that.
The use case is being able to install multiple versions of the same ext and something we had a lot of feedback from in Desktop when we launched with the same limitations.

Maybe out of scope for this PR though.

packages/common/src/utils/extensions/extension-versions.ts
nglgzz: Ah right, I missed that part, I thought it was just the response coming from npm.

nglgzz (dismissed):
Looks good to me. Looking forward to seeing Relate extensions on NPM :sparkles:

jk05 (dismissed):

package.json
```diff
@@ -16,7 +16,7 @@
         "lint": "eslint . --ext .ts",
         "test:teardown": "node scripts/tests/teardown.js",
         "test:setup": "node scripts/tests/setup.js",
-        "lerna:test": "lerna run test --concurrency 1",
+        "lerna:test": "lerna run test",
```
nglgzz, line 19: I'm afraid this will increase flakiness. Not because our tests can't run concurrently (they should) but because of the intensive file operations in our tests.

packages/common/src/entities/dbmss/dbmss.local.ts
```diff
@@ -122,8 +122,13 @@ export class LocalDbmss extends DbmssAbstract<LocalEnvironment> {
 
         // version as a file path.
         if ((await fse.pathExists(version)) && (await fse.stat(version)).isFile()) {
-            const {extractedDistPath} = await extractNeo4j(version, this.environment.dirPaths.dbmssCache);
-            return this.installNeo4j(name, this.getDbmsRootPath(), extractedDistPath, credentials);
+            const tmpPath = path.join(this.environment.dirPaths.tmp, uuidv4());
+            const {extractedDistPath} = await extractNeo4j(version, tmpPath);
+            const dbms = await this.installNeo4j(name, this.getDbmsRootPath(), extractedDistPath, credentials);
+
+            await fse.remove(tmpPath);
```
nglgzz, line 129: :+1:

package.json
huboneo: I just ran 10 builds in a row without failures, so I think it's ok? we shouldnt be manipulating the same file at the same time?

nglgzz (approved):

oskarhane:
> My one concern here is that we check the signature of an extension every time it is discovered (i.e. listed). This enables us to catch revocations on the fly, but slows down list operations due to network latency. WDYT @oskarhane ?

Good call. Let's start like this and optimize in the future if needed.

PR #219, opened 2020-10-27
common/extensions install and web/extensions.e2e tests

jk05:

Replica of https://github.com/neo4j-devtools/relate/pull/152 (was created before the repo switch over and it seems to have lost context)
- have had to add filtering to check the the test extensions names include `path.basename(__filename)`, for example, as there didnt seem to be a guarantee of what can be in fixtures directory at any moment in time - was causing a lots of flaky test failures on builds otherwise.

### What is the new behavior?
(if this is a feature change)

### DIFF
--- a/packages/common/src/utils/system/test-extensions.ts
+++ b/packages/common/src/utils/system/test-extensions.ts
@@ -5,7 +5,7 @@ import tar from 'tar';
 
 import {envPaths} from '../env-paths';
 import {IInstalledExtension} from '../../models';
-import {EXTENSION_DIR_NAME, EXTENSION_MANIFEST_FILE, EXTENSION_TYPES} from '../../constants';
+import {ENTITY_TYPES, EXTENSION_DIR_NAME, EXTENSION_MANIFEST_FILE, EXTENSION_TYPES} from '../../constants';
 
 export class TestExtensions {
     extensions: Array<IInstalledExtension | string> = [];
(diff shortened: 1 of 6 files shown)


jk05:
Im not entirely convinced about the stability of these tests...

packages/common/src/utils/system/test-extensions.ts
```diff
@@ -74,34 +77,15 @@ export class TestExtensions {
...
-                    path.join(envPaths().cache, EXTENSION_DIR_NAME, ext.name),
-                    path.join(envPaths().cache, EXTENSION_DIR_NAME, `${ext.name}@${ext.version}`),
-                    path.join(envPaths().data, EXTENSION_DIR_NAME, ext.name),
-                    path.join(envPaths().data, EXTENSION_DIR_NAME, `${ext.name}@${ext.version}`),
-                ].map((p) => fse.remove(p)),
-            );
-        });
-
-        await Promise.all(deleteAll);
+        await fse.emptyDir(path.join(envPaths().cache, EXTENSION_DIR_NAME));
```
jk05, line 88: I did this initially for other reasons and was going to revert, but looking at it again, if it works would it be ok to leave it like this in the teardown?

nglgzz: The point of this method is to delete only the extensions installed by a specific TestExtensions instance. If we delete all the content of those directories we won't be able to run tests concurrently, which might lead to more flakiness in the future. 

Is the previous implementation not working as expected?

jk05: Thats a good point, I will put it back. Its been a while since I did this work so I cant recall. It probably works fine, Ill put it back and try again.

nglgzz (dismissed):
Great work ✨

nglgzz (approved):

PR #223, opened 2020-10-29
Feature remote STATIC extensions

huboneo:

This PR adds support for using remote apps, defined as an installed STATIC extension whose `main` property contains a URL.

This PR also updates various URL parameters used by relate to match those of Neo4j Desktop, and exposes a new client utility for grabbing them from URLs.

Further this PR changes the GraphQL schema for apps to include a `url` property instead of a `path`.

Can only open locally installed apps

### What is the new behavior?
Apps can point to a URL

yes, this PR changes the GraphQL schema for apps to include a `url` property instead of a `path`.

### Other information:
Sample remote app:
```json
{
  "name": "gewgle",
  "version": "1.0.1",
  "main": "https://google.com"
}
```
[extension-gewgle.zip](https://github.com/neo4j-devtools/relate/files/5458309/extension-gewgle.zip)

CLI Usage
```
$ relate app:open gewgle -L
> https://google.com/?relateUrl=http%3A%2F%2F127.0.0.1%3A3000
```

WEB Usage
Screenshot 2020-10-29 at 13 32 57

### DIFF
--- a/packages/common/src/utils/extensions/get-app-launch-url.ts
+++ b/packages/common/src/utils/extensions/get-app-launch-url.ts
@@ -1,5 +1,6 @@
 import {EnvironmentAbstract} from '../../entities/environments';
 import {isValidUrl} from '../generic';
+import {RELATE_API_TOKEN_PARAM_NAME, RELATE_LAUNCH_TOKEN_PARAM_NAME, RELATE_URL_PARAM_NAME} from '../../constants';
 
 export async function getAppLaunchUrl(
     environment: EnvironmentAbstract,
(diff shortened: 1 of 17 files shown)


packages/common/src/utils/extensions/get-app-launch-url.ts
```diff
@@ -0,0 +1,25 @@
...
+    launchToken?: string,
+): Promise<string> {
+    const asURL = isValidUrl(appPath) ? new URL(appPath) : new URL(`${environment.httpOrigin}${appPath}`);
+
+    if (asURL.origin !== environment.httpOrigin) {
+        asURL.searchParams.set('relateUrl', environment.httpOrigin);
+    }
+
+    if (environment.requiresAPIToken) {
+        asURL.searchParams.set('relateApiToken', await environment.generateAPIToken(asURL.host, clientId, {}));
```
huboneo, line 17: There is an assumption here, that appName is essentially the clientID. I dont know how we could enforce/communicate this convention to outside devs

packages/common/src/utils/extensions/get-app-launch-url.ts
```diff
@@ -0,0 +1,25 @@
...
+export async function getAppLaunchUrl(
+    environment: EnvironmentAbstract,
+    appPath: string,
+    clientId: string,
+    launchToken?: string,
+): Promise<string> {
+    const asURL = isValidUrl(appPath) ? new URL(appPath) : new URL(`${environment.httpOrigin}${appPath}`);
+
+    if (asURL.origin !== environment.httpOrigin) {
+        asURL.searchParams.set('relateUrl', environment.httpOrigin);
```
nglgzz, line 13: Shouldn't we use constants for these parameters?

huboneo: I'll add it in, but not sure who would use them?

nglgzz (approved):
Looks good to me

packages/common/src/utils/extensions/get-app-launch-url.ts
huboneo: but yeah, good call :)

PR #226, opened 2020-10-29
Chore separate access token files

huboneo:

This PR saves access tokens to separate files to avoid race-conditions when multiple processes are trying to create them at the same time

Concurrent tests fail due to overwrites

### What is the new behavior?
Concurrent tests pass

### DIFF
--- a/packages/common/src/utils/system-access-token.ts
+++ b/packages/common/src/utils/system-access-token.ts
@@ -19,13 +19,18 @@ export async function registerSystemAccessToken(
     await fse.writeFile(path.join(tokenDirPath, fileName), accessToken, 'utf8');
 }
 
-export function getSystemAccessToken(
+export async function getSystemAccessToken(
     tokenDirPath: string,
     environmentNameOrId: string,
     dbmsId: string,
     dbmsUser: string,
 ): Promise<string | undefined> {
     const fileName = getAccessTokenFileName(environmentNameOrId, dbmsId, dbmsUser);
+    const filePath = path.join(tokenDirPath, fileName);
 
-    return fse.readFile(path.join(tokenDirPath, fileName), 'utf8').catch(() => undefined);
+    if (await fse.pathExists(filePath)) {
+        return fse.readFile(filePath, 'utf8');
+    }
+
+    return undefined;
 }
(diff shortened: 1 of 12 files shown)


packages/common/src/utils/system-access-token.ts
```diff
@@ -1,32 +1,31 @@
-import {PropertiesFile} from '../system/files';
+import fse from 'fs-extra';
+import path from 'path';
 
-export function getAccessTokenRCKey(environmentNameOrId: string, dbmsId: string, dbmsUser: string): string {
-    return `//${environmentNameOrId}/${dbmsId}/${dbmsUser}/:_accessToken`;
+import {ENTITY_TYPES} from '../constants';
+
+export function getAccessTokenFileName(environmentNameOrId: string, dbmsId: string, dbmsUser: string): string {
+    return `${environmentNameOrId}-${ENTITY_TYPES.DBMS}-${dbmsId}-${dbmsUser}`;
```
nglgzz, line 7: We don't actually want names here, right? 

```suggestion
export function getAccessTokenFileName(environmentId: string, dbmsId: string, dbmsUser: string): string {
    return `${environmentId}-${ENTITY_TYPES.DBMS}-${dbmsId}-${dbmsUser}`;
```

packages/common/src/utils/system-access-token.ts
```diff
@@ -1,32 +1,31 @@
...
     environmentNameOrId: string,
     dbmsId: string,
     dbmsUser: string,
 ): Promise<string | undefined> {
-    const key = getAccessTokenRCKey(environmentNameOrId, dbmsId, dbmsUser);
-    const properties = await PropertiesFile.readFile(knownConnectionsPath);
+    const fileName = getAccessTokenFileName(environmentNameOrId, dbmsId, dbmsUser);
 
-    return properties.get(key);
+    return fse.readFile(path.join(tokenDirPath, fileName), 'utf8').catch(() => undefined);
```
nglgzz, line 30: Is the catch meant to handle the case where the file doesn't exist? If yes, could we throw for everything else (or do a check beforehand)? 

I'm thinking if we get some permission error we should bubble up the error to the user instead of defaulting to undefined and pretend everything is fine.

nglgzz (commented):
Awesome work finding this issue!! Just a couple of minor comments.

nglgzz (approved):

huboneo:
@nglgzz please review again :)

nglgzz (approved):

huboneo:
@nglgzz one more time :D i updated the linter to ban type assertions

nglgzz (approved):

PR #229, opened 2020-11-09
create separate env dirs when creating new envs

jk05:

- New environments being set up will have separate data directories created (`environment-{id}`) under the Relate data dir
- Add static `register` method to `SystemModule`, `ElectronModule` and `WebModule` to allow consuming modules to optionally create dynamic modules by passing in config (to get the environment for extension loading purposes).
- `systemProvider.getEnvironment` will now check for `nameOrId`, then config service `defaultEnvironmentNameOrId` ands finally an active environment.
- cleanup of specifically `envPaths.data()` to use environment `dataPath` where appropriate.
- `test-extensions` to use correct paths

- Environments all share the same Relate data directory, which can be unexpected.
- `relateDataPath` can be passed to the config to set a new data path.

### What is the new behavior?
- New  environments created will have `relateDataPath` set as separate directories (`environment-{id}`) under Relate global data dir. Existing envs will remain as is (sharing the global Relate data dir).

`relateDataPath` is not new and does allow creating data directories to a different path. However, there could be some issues.


packages/common/src/system/system.provider.ts
```diff
@@ -112,6 +112,8 @@ export class SystemProvider implements OnModuleInit {
         const newId = uuidv4();
         const fileName = `${config.name}${JSON_FILE_EXTENSION}`;
         const filePath = path.join(this.dirPaths.environmentsConfig, fileName);
+        const dataDir = config.relateDataPath || this.dirPaths.data;
+        const relateDataPath = path.join(dataDir, `${ENTITY_TYPES.ENVIRONMENT}-${config.name}`);
```
huboneo, line 116: We do not use names in filepaths by convention, use ID instead?

packages/common/src/utils/extensions/get-installed-extensions.ts
```diff
@@ -20,8 +20,9 @@ import {
 /**
  * Synchronous method, only call on process bootstrap (not inside applications)
  */
-export function getInstalledExtensionsSync(): List<IInstalledExtension> {
-    const installationDir = path.join(envPaths().data, EXTENSION_DIR_NAME);
+export function getInstalledExtensionsSync(installDataDir?: string): List<IInstalledExtension> {
```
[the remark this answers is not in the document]
jk05: ultimately don't want this to be optional and want to use the environment data extension directory (not `envPaths().data`). I did this as an interim step until I get ideas of how to deal with the other places its called and not immediately obvious what to do. As for if it would work without the `dataDir`, I dont think it can as it stands currently.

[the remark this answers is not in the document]
huboneo: I say make it mandatory and see if typescript complains, should be enough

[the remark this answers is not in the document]
nglgzz: I think the problem here is that `getInstalledExtensionsSync` is used in `loadExtensionsFor` and that function is used in a few places where we don't have access to an environment.

packages/common/src/entities/environments/environment.abstract.ts
```diff
@@ -91,7 +91,7 @@ export abstract class EnvironmentAbstract {
      * @hidden
      */
     public get dataPath(): string {
-        return this.config.relateDataPath || path.join(envPaths().data, `${ENTITY_TYPES.ENVIRONMENT}-${this.id}`);
+        return path.join(this.config.relateDataPath || envPaths().data, `${ENTITY_TYPES.ENVIRONMENT}-${this.id}`);
```
huboneo, line 94: @jk05 here is the root of the issues we were seeing

huboneo:
@jk05 @nglgzz I pushed a fix for the failing tests. Basically from now on we have to use the environment paths (not envPath()) to find various directories

packages/common/src/entities/environments/environment.abstract.ts
nglgzz: So if an environment doesn't have a `relateDataPath` it will default to a separate folder in the data directory?

jk05: ah right!

huboneo: Well nomatter it will create a separate folder in the data directory

huboneo: You could create several environments in the same data path, the data path is for the entire relate system so to speak (including acceptedTerms)

huboneo: But maybe that's not what we want?

nglgzz: This is more breaking than it's mentioned in the description then. Because it's not "not migrating old environments", it's "pointing them to a new location without moving their data".

huboneo: " meaning any already created environments would need to be migrated manually"?

nglgzz: The way I thought about it was, if you don't specify anything in the config, it will default to the global data folder, otherwise it will use relateDataPath, and relateDataPath will be set automatically for newly created environments.

huboneo: thats a confusion default if you ask me. Wouldn't it be better to just change this behaviour now while still in alpha?

huboneo: I'd be open to either, but right now relate does not seem like a happy bunny if I try to change this property to anything else

nglgzz: Migrating the data is out of scope for this feature, and if we leave it out here, we have to add it into Desktop.

>  meaning any already created environments would need to be migrated manually

I'm fine with this meaning that old environments are shared unless you migrate them, but not that you won't see the data unless you migrate the environments.

nglgzz: @huboneo The defaults for the user are the same, it's just that by setting the default on  `systemProvider.createEnvironment` instead of here you can have this change be backward compatible.

https://github.com/neo4j-devtools/relate/blob/be2a532bf5080c5297d644e9db095a138c5686d1/packages/common/src/system/system.provider.ts#L128-L132

packages/common/src/system/system.provider.ts
```diff
@@ -110,6 +139,7 @@ export class SystemProvider implements OnModuleInit {
         }
 
         const configModel = new EnvironmentConfigModel({
+            relateDataPath: path.join(envPaths().data, `${ENTITY_TYPES.ENVIRONMENT}-${newId}`),
```
jk05, line 142: `path.join(config.relateDataPath || envPaths().data, `${ENTITY_TYPES.ENVIRONMENT}-${newId}`)` is more along the lines of what I thought we would want.

jk05: Spoke with @nglgzz , this is fine as it is.

huboneo (commented):
Looks good so far. Could you add a test for system provider where it uses a non-active environment as default?

huboneo (approved):
LGTM

jk05:
[[Kick-off] Place Relate data inside Desktop folder](https://trello.com/c/mkoONDHP/1497-kick-off-place-relate-data-inside-desktop-folder)

nglgzz:
[local dbmss are now shared between all local envs. Is this good or bad?](https://trello.com/c/A1AZkgpt/149-local-dbmss-are-now-shared-between-all-local-envs-is-this-good-or-bad)

PR #232, opened 2020-11-16
Add metadata field to entity manifests

nglgzz:

Feature

- There is no way to store extra information about entities.
- There is no way to do a shallow merge on a Dict. 

### What is the new behavior?
- Extra information about entities can be stored in the metadata field of their manifest. 
- The Dict type has a new `assign` method, equivalent to `Object.assign`.

No


packages/common/src/entities/manifest/manifest.local.ts
```diff
@@ -23,6 +23,35 @@ export class ManifestLocal<Entity extends IManifest, Manifest extends ManifestMo
...
+                .toObject(),
+        });
+
+        return this.getEntity(id);
+    }
+
+    public async removeMetadata(nameOrId: string, keys: string[]): Promise<Entity> {
+        const {id, metadata} = await this.getEntity(nameOrId);
+
+        const updated = Dict.from(metadata)
```
huboneo, line 43: this is a great usecase for `Dict.omit()` otherwise, but I realise now I implemented it incorrectly. Omit should be: 
```typescript
omit<K2 extends K = K, R = T extends object ? Omit<T, K2> : never>(...other: K2[]): Dict<R>;

omit<K2 extends KeyVal<T>['key'], I = T extends Map<K, V> ? T : never, R = Dict<T, Exclude<K, K2>, V>>(
    ...other: K2[],
): R;

/**
 * Omits one or more keys from the Dict
 * ```ts
 * const fooBar = Dict.from({foo: true, bar: 1});
 * const foo = fooBar.omit('bar');
 * foo.toObject() // {foo: true}
 * ```
 */
omit(...other: any[]) {
    return this.toList()
        .filter(([key]) => !other.includes(key)
        .switchMap(Dict.from);
}
```
Not sure it the built in `Omit<>` generic can handle this syntax tho

huboneo (dismissed):
LGTM

packages/web/src/entities/project/project.types.ts
```diff
@@ -5,7 +5,7 @@ import {IFileUpload, GraphQLUpload} from './graphql-upload';
 import {EnvironmentArgs, RelateFile} from '../../global.types';
 
 @ObjectType()
-export class Project implements Omit<IProject, 'root'> {
+export class Project implements Omit<IProject, 'root' | 'metadata'> {
```
nglgzz, line 8: Not entirely sure what type to use for metadata on GraphQL, so I omitted it for now. Suggestions are welcome.

huboneo: https://www.npmjs.com/package/graphql-type-json

huboneo: alternatively you can make your own object type, as we do not support deeply nested objects(?)

packages/common/src/entities/manifest/manifest.local.ts
nglgzz: I actually thought about it but was too lazy to deal with the types. That snippet seems to work though, so I'll just add it.

packages/web/src/entities/project/project.types.ts
nglgzz: That scalar should be exactly what we need, will add it on a separate PR.

huboneo (approved):
ღゝ◡╹)ノ♡

huboneo (approved):

nglgzz:
[[Kick-off] Link DBMSs to projects with Relate in Desktop](https://trello.com/c/fIXpxVW1/1498-kick-off-link-dbmss-to-projects-with-relate-in-desktop)

PR #234, opened 2020-11-18
add is-symlink check and update project and dbms link logic

jk05:

- add `is-symlink` util
- update `dbms.local` and `project.local` `link` methods to `unlink` if already a symlink. This is to allow relinking broken symlinks.
- updated `project` link to be in line with `dbms` link

If a dbms or project is already linked, and the target is moved, the symlink will remain and is broken. Project link also doesnt handle linking a previously linked Project (already has a manifest) in the same way as dbms

### What is the new behavior?
- When linking for project or dbms, when it already has a manifest and so previously linked, the presence of a symlink is checked and unlinked first. 
- Project link in line with dbms link now.

No, other than a discussion around a flag to prevent removing 'broken' links in the case when projects, for example, are loaded via storage device and not always available. This is best handled when that work is done in future to keep the scope focus here.

### DIFF
--- a/packages/common/src/entities/dbmss/dbmss.local.ts
+++ b/packages/common/src/entities/dbmss/dbmss.local.ts
@@ -65,6 +65,7 @@ import {PropertiesFile} from '../../system/files';
 import {winNeo4jStart, winNeo4jStatus, winNeo4jStop} from '../../utils/dbmss/neo4j-process-win';
 import {emitHookEvent} from '../../utils';
 import {ManifestLocal} from '../manifest';
+import {isSymlink} from '../../utils/files';
 
 export class LocalDbmss extends DbmssAbstract<LocalEnvironment> {
     public readonly manifest = new ManifestLocal(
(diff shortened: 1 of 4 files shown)


packages/common/src/entities/dbmss/dbmss.local.ts
```diff
@@ -315,6 +316,9 @@ export class LocalDbmss extends DbmssAbstract<LocalEnvironment> {
                 throw new InvalidArgumentError(`DBMS "${name}" already managed by relate`);
             }
 
+            if (await isSymlink(target)) {
```
nglgzz, line 319: I would add a comment explaining that we can get here only when a symlink is broken. It might not be straightforward that the `targetExists` check above returns true only if the symlink is pointing to an existing path.

nglgzz (dismissed):
Looks good to me

nglgzz (approved):

jk05:
[[Kick-off] Place Relate data inside Desktop folder](https://trello.com/c/mkoONDHP/1497-kick-off-place-relate-data-inside-desktop-folder)

PR #241, opened 2020-12-08
Recursive merge on manifest update, swallow errors on manifest reads

nglgzz:

Optimization + reintroducing old behavior

- `manifest.update` uses `_.assign` to assign new properties.
- Errors thrown while reading entity manifests are not handled.

### What is the new behavior?
- `manifest.update` uses `_.merge` to set new properties (by default). This allows adding new tags/metadata, without overriding previous values.
- Errors thrown while reading entity manifests are swallowed and default values are returned in those cases.
- Hooks are triggered right before reading/writing to manifests (these were added for debug purposes).

No

### Other information:
Reintroducing the old behavior for reading manifests, is a temporary "fix" until we get more time to fix the race condition when manifests are read and written at the same time.


packages/common/src/entities/manifest/manifest.local.ts
```diff
@@ -97,21 +110,33 @@ export class ManifestLocal<Entity extends IManifest, Manifest extends ManifestMo
...
-        const manifest = await fse.readJSON(manifestPath, {encoding: 'utf-8'});
-        return new this.EntityModel({
-            ...defaults,
-            ...manifest,
-            id,
-        });
+        // @todo - We should bubble up the error here, but to do so we need to find a
+        // reliable way to ensure we're not reading while the file is being written.
+        try {
+            await emitHookEvent(HOOK_EVENTS.MANIFEST_READ, manifestPath);
```
jk05, line 116: 👍

packages/common/src/entities/manifest/manifest.local.ts
```diff
@@ -97,21 +110,33 @@ export class ManifestLocal<Entity extends IManifest, Manifest extends ManifestMo
...
+            const manifest = await fse.readJSON(manifestPath, {encoding: 'utf-8'});
+
+            return new this.EntityModel({
+                ...defaults,
+                ...manifest,
+                id,
+            });
+        } catch (e) {
+            return new this.EntityModel({
```
jk05, line 125: will this cause any issues further on by just returning defaults? or should it be ok?

huboneo: Yeah not sure this method should hide a missing manifest, maybe better to do it case by case where `get()` gets called

nglgzz: This was the previous behavior and it is (and was) causing problems. It's just that right now it's easier to leave it as is than addressing the race condition I've been investigating. Wouldn't want to rush a solution that doesn't actually work reliably.

nglgzz: Returning default values on a missing manifest is intended, and that's not causing any issues. The problem is with reading a manifest while it's being written/updated. This is the same as what we currently have in production:

https://github.com/neo4j-devtools/relate/blob/20b649d9d7dd9c752fc71fd18f4af0fae28ca61b/packages/common/src/entities/dbmss/dbmss.local.ts#L762-L775

jk05 (dismissed):
@nglgzz will be exploring a 'proper' solution after the dev phase, so happy to approve this now to unblock

jk05 (approved):

nglgzz:
[Canary sometimes fails to start with "Unexpected end of JSON input"](https://trello.com/c/puoT8P0y/1530-canary-sometimes-fails-to-start-with-unexpected-end-of-json-input)

PR #245, opened 2020-12-14
Install sample projects

linuslundahl:

It introduces a new command `relate project:install-sample`, which grabs a list of Sample Projects from https://github.com/neo4j-graph-examples that are marked as `neo4j-approved` or some other filter that we decide on.
These projects contains sample data and also a new file (`relate.project-install.json`) that describes how Relate/Desktop should install Sample DBMSs and import sample data. See https://github.com/neo4j-graph-examples/northwind for an example.
The user will be prompted to install any DBMSs found under the dbms section in the file.

Note!

Currently the file is named `relate.project.json` in the northwind example. Because of this the second step doesn't work when testing the command.

### What is the new behavior?
(if this is a feature change)

### DIFF
--- a/packages/common/src/entities/environments/environment.local.ts
+++ b/packages/common/src/entities/environments/environment.local.ts
@@ -56,6 +56,9 @@ export class LocalEnvironment extends EnvironmentAbstract {
             case ENTITY_TYPES.PROJECT:
                 return path.join(this.dirPaths.projectsData, `${ENTITY_TYPES.PROJECT}-${id}`);
 
+            case ENTITY_TYPES.PROJECT_INSTALL:
+                return path.join(this.dirPaths.projectsData, `${ENTITY_TYPES.PROJECT}-${id}`);
+
             case ENTITY_TYPES.EXTENSION:
                 return path.join(this.dirPaths.extensionsData, `${ENTITY_TYPES.EXTENSION}-${id}`);
 
--- a/packages/common/src/entities/projects/projects.local.ts
+++ b/packages/common/src/entities/projects/projects.local.ts
@@ -1,24 +1,47 @@
 import path from 'path';
 import fse from 'fs-extra';
 import {from} from 'rxjs';
+import semver from 'semver';
 import {flatMap, map} from 'rxjs/operators';
-import {List, Maybe, None, Str} from '@relate/types';
 import {v4 as uuidv4} from 'uuid';
+import got, {Got} from 'got';
+import {List, Maybe, None, Str} from '@relate/types';
 
-import {IProject, IProjectDbms, IProjectInput, IRelateFile, ProjectManifestModel, WriteFileFlag} from '../../models';
+import {
+    IProject,
+    IProjectDbms,
+    IProjectInput,
+    IRelateFile,
+    ProjectManifestModel,
+    ProjectInstallManifestModel,
+    WriteFileFlag,
+    ISampleProjectDbms,
+    ISampleProjectInput,
+    IDbmsInfo,
+    IDbmsVersion,
+    ISampleProjectRest,
+} from '../../models';
 import {ProjectsAbstract} from './projects.abstract';
 import {LocalEnvironment} from '../environments';
 import {ManifestLocal} from '../manifest';
 import {ENTITY_TYPES, PROJECT_MANIFEST_FILE} from '../../constants';
-import {InvalidArgumentError, NotFoundError} from '../../errors';
+import {FetchError, InvalidArgumentError, NotFoundError} from '../../errors';
 import {getAbsoluteProjectPath, getRelativeProjectPath, isSymlink, mapFileToModel} from '../../utils/files';
 import {applyEntityFilters, IRelateFilter} from '../../utils/generic';
+import {download, extract} from '../../utils';
 
 export class LocalProjects extends ProjectsAbstract<LocalEnvironment> {
     public readonly manifest = new ManifestLocal(
         this.environment,
         ENTITY_TYPES.PROJECT,
         ProjectManifestModel,
+        (nameOrId: string) => this.get(nameOrId),
+    );
+
+    public readonly manifestSampleProject = new ManifestLocal(
+        this.environment,
+        ENTITY_TYPES.PROJECT_INSTALL,
+        ProjectInstallManifestModel,
         (nameOrId: string) => this.get(nameOrId),
     );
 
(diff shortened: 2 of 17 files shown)


packages/common/src/entities/projects/projects.local.ts
```diff
@@ -22,6 +39,25 @@ export class LocalProjects extends ProjectsAbstract<LocalEnvironment> {
...
+    );
+
+    private graphqlClient: GraphQLAbstract;
+
+    constructor(config: any) {
+        super(config);
+
+        // @TODO: Check how the token should be stored/generated.
+        this.graphqlClient = new GraphQLClient({
+            headers: {Authorization: 'bearer 796553eeddc5e27223a8dac0f90131fe8cf3f7c9'},
```
linuslundahl, line 56: This Token should probably be stored somewhere else and also not be generated by me? I'm not sure what we should do here, but the GitHub API requires a Token even if the organisation and the repositories are public.

linuslundahl: This token is now revoked.

nglgzz: The REST API doesn't require authentication. Would it make sense to use that one instead?

linuslundahl: In theory it could I guess, but it would require many more requests to do what is done with one request via GraphQL... I think we can add the token via the build-step instead. The token we use will of course be in the built version, but like I said on standup, it won't give access to anything other than reading public repos.

nglgzz: If we used a prefix in the description instead of (or in addition to) topics we would only need one request: https://api.github.com/orgs/neo4j-graph-examples/repos

linuslundahl: I think I have solved it. It will require developers to add an `.env` file locally with a token though for test to run and the `install-sample` command to work.
`echo 'GITHUB_API_TOKEN=${TOKEN_FROM_GITHUB}' > .env`

nglgzz: I feel like this is adding unnecessary complexity and a broken experience, especially considering that we only need this to list public repositories, and that is available with one REST request and no token. 

For the filtering, sure it would require more requests if we wanted to use topics. But at the moment there are only two approved repos, so I doubt it would be a problem to change the filtering to be based on something else, so we can keep this solution as simple as possible.

linuslundahl: Yeah, fair point.

huboneo:
Just wanted to say that the code looks really good! Feels like relate has matured a lot and made it easy to add functionality, even if you're not a core author(?)

linuslundahl:
> Just wanted to say that the code looks really good! Feels like relate has matured a lot and made it easy to add functionality, even if you're not a core author(?)

Since I have already been poking around here I'm not completely new to it, but it's certainly well structured and easily navigated!

packages/common/src/models/project-install.model.ts
```diff
@@ -0,0 +1,52 @@
...
+            };
+        }>;
+    };
+}
+
+export class ProjectInstallManifestModel extends ManifestModel<ISampleProject> implements ISampleProjectManifest {
+    constructor(props: any) {
+        super(props);
+
+        // reassigning default values doesn't work when it's done from the parent class
```
huboneo, line 45: @nglgzz this is worth looking in to

huboneo:
Just checking that `relate.project-install.json` in the PR description is a typo, should be `relate.project.json` right?

linuslundahl:
> Just checking that `relate.project-install.json` in the PR description is a typo, should be `relate.project.json` right?

Well, we have discussed this back and forth, `relate.project.json` is a state of what the project looks like on disk, `relate.project-install.json` is a new file that describes instructions for how to install a project, since files on disk might be removed etc there's no reason to have those instructions saved in the `relate.project.json`-file.

huboneo:
Ok, just asking as the examples currently contain `relate.project.json` files

linuslundahl:
> Ok, just asking as the examples currently contain `relate.project.json` files

Yes, correct. And that should be changed. :)

packages/common/src/models/project-install.model.ts
nglgzz: Yeah I started adding these. I tried to look into it when I was doing the manifest changes, but couldn't find a way around this that doesn't involve an extra dependency. 

The issue is that when setting a default value for a property in the model, the parent constructor is not overwriting it. To reproduce the issue you can run this snippet: 

```javascript
class Foo { 
  constructor(props) { 
    Object.assign(this, props) 
  } 
}
class Bar extends Foo { a = 10 }

const bar = new Bar({a: 20}) // Bar { a: 10 }
```

packages/common/src/entities/environments/environment.local.ts
```diff
@@ -56,6 +56,9 @@ export class LocalEnvironment extends EnvironmentAbstract {
             case ENTITY_TYPES.PROJECT:
                 return path.join(this.dirPaths.projectsData, `${ENTITY_TYPES.PROJECT}-${id}`);
 
+            case ENTITY_TYPES.PROJECT_INSTALL:
+                return path.join(this.dirPaths.projectsData, `${ENTITY_TYPES.PROJECT}-${id}`);
```
nglgzz, line 60: I'm guessing this is meant to be:

```suggestion
                return path.join(this.dirPaths.projectsData, `${ENTITY_TYPES.PROJECT_INSTALL}-${id}`);
```

packages/common/package.json
```diff
@@ -72,6 +72,7 @@
         "apollo-link-http": "1.5.17",
         "class-validator": "0.12.1",
         "decompress": "4.2.1",
+        "dotenv": "8.2.0",
```
nglgzz, line 75: We don't need this change anymore (same with package-lock).

packages/common/src/utils/graphql.ts
```diff
@@ -0,0 +1,70 @@
...
+     * @hidden
+     */
+    constructor(protected readonly config: IGraphQLConfig) {}
+
+    /**
+     * Executes an operation
+     * @operation   operation
+     */
+    abstract execute(operation: GraphQLRequest): Promise<FetchResult<{[key: string]: any}>>;
+}
```
nglgzz, line 20: Not sure if we really need the abstract class in this case. I doubt we'll be swapping Apollo client with other implementations.

packages/common/src/entities/projects/projects.local.ts
```diff
@@ -22,6 +38,17 @@ export class LocalProjects extends ProjectsAbstract<LocalEnvironment> {
...
+    public readonly manifestSampleProject = new ManifestLocal(
+        this.environment,
+        ENTITY_TYPES.PROJECT_INSTALL,
+        ProjectInstallManifestModel,
+        (nameOrId: string) => this.get(nameOrId),
+    );
+
+    constructor(config: any) {
+        super(config);
+    }
```
nglgzz, line 50: Is the constructor really needed?

packages/common/src/entities/projects/projects.local.ts
```diff
@@ -267,6 +294,136 @@ export class LocalProjects extends ProjectsAbstract<LocalEnvironment> {
...
+        } catch (_error) {
+            throw new FetchError(`Unable to fetch Sample Projects from GitHub`);
+        }
+    }
+
+    async downloadSampleProject(selected: string, destPath?: string): Promise<{path: string; temp: boolean}> {
+        const {tmp} = this.environment.dirPaths;
+
+        const diskPath = destPath || path.join(tmp, 'sample-project');
+        const sampleProject = await download(
```
nglgzz, line 320: It might be worth trying to delete the file if it already exists.

packages/common/src/entities/projects/projects.local.ts
```diff
@@ -267,6 +294,136 @@ export class LocalProjects extends ProjectsAbstract<LocalEnvironment> {
...
+            });
+        }
+
+        if (temp) {
+            await fse.copy(srcPath, project.root, {filter: (src) => !src.includes('relate.project.json')});
+        }
+
+        let install;
+        try {
+            install = await this.manifestSampleProject.get(project.id);
```
nglgzz, line 364: If the manifest doesn't exist this will just return the default values for it, so the try/catch is not needed.

packages/common/src/entities/projects/projects.local.ts
```diff
@@ -267,6 +294,136 @@ export class LocalProjects extends ProjectsAbstract<LocalEnvironment> {
...
+            temp: !destPath,
+        };
+    }
+
+    async installSampleProject(
+        srcPath: string,
+        args: {
+            name?: string;
+            nameOrId?: string;
+            temp?: boolean;
```
nglgzz, line 338: What's the difference between `name` and `nameOrId`?

packages/common/src/entities/projects/projects.local.ts
```diff
@@ -267,6 +294,136 @@ export class LocalProjects extends ProjectsAbstract<LocalEnvironment> {
...
+
+        let project;
+
+        if (nameOrId && !temp) {
+            project = await this.get(nameOrId);
+        } else if (!nameOrId && !temp && name) {
+            project = await this.link(srcPath, name);
+        }
+
+        if (!project) {
```
nglgzz, line 351: Should we use the existing project if one already exists, or should we just throw an error? I'm guessing if a user is trying to install a sample project they probably don't want to overwrite an existing one.

nglgzz (commented):
As Hugo mentioned overall looks good, just some details/questions.

packages/common/src/models/project-install.model.ts
huboneo: Ah, yes that is an inherent limitation of the syntactic sugar that is JS classes. It is assigned, then overwritten by super

packages/common/src/entities/environments/environment.local.ts
linuslundahl: No, I made that mistake too. But the project id that a relate.project-install.json is located in is that one.

packages/common/package.json
linuslundahl: 👍

linuslundahl: To my defence I actually ran `npm uninstall dotenv`... but in the wrong directory...

packages/common/src/utils/graphql.ts
linuslundahl: This is a leftover from when we used GraphQL instead to fetch from GitHub. This makes it simpler to fetch via GraphQL in the future if we need it, but if we feel it's unnecessary I can remove it.

packages/common/src/entities/projects/projects.local.ts
linuslundahl: Also leftover from GraphQL. Will remove.

packages/common/src/entities/projects/projects.local.ts
linuslundahl: Well, in the temp folder it doesn't matter... if the user provides a path where it should download we don't want to remove any existing files. In that case I'm actually not sure if the download function replaces it or adds a suffix or just fails with a `File already exists`. Worth investigating.

packages/common/src/entities/projects/projects.local.ts
linuslundahl: Well, I guess it can be `name` and `id`. Just used the naming convention of existing vars.

packages/common/src/entities/projects/projects.local.ts
linuslundahl: Well, there are three scenarios here.

1. `projectId` is provided, we just get the project info from the id and return. Desktop does this.
2. A `name` and a `srcPath` exists, this is a folder on disk that should be linked. 
3. None of the above is satisfied, we create a new project from scratch and copy the files there.

packages/common/src/entities/projects/projects.local.ts
linuslundahl: 👍 had another solution within the try/catch before, going to blame it on that. 😉

oskarhane (approved):
It's good to see how clean it is to add features to this codebase.
Looks good now 👍

packages/common/src/entities/projects/projects.local.ts
nglgzz: Thinking again about this `download` will overwrite whatever path you pass it. And since the generated path is always the same, this can cause race conditions. For example, if you download two different samples without specifying a location, and then install them, you'll end up installing the last downloaded sample twice.

huboneo: typically we use `uuid()` to solve this

linuslundahl: Sure, using a random name for the download makes sense. I'll add that.

packages/common/src/entities/projects/projects.local.ts
nglgzz: What is `srcPath`? Is it the path to the extracted sample project or the path where the project should be installed? 

Reading it again, I think I might have misunderstood what this method is doing, it seems like this is not actually installing the sample project, but rather getting/creating/linking the project where the data will be copied and where the DBMSs will be created (?)

I'm still not sure why we would expect to use an existing project when installing a sample project though. Is it just because Desktop is creating it instead?

huboneo: sounds like a great test case?

packages/common/src/entities/projects/projects.local.ts
huboneo: sounds like a great test case?

PR #246, opened 2020-12-16
RFC dbms plugins

huboneo:

This PR adds an RFC for DBMS plugins, as well as declaring extension dependencies

See [RFC doc](https://github.com/huboneo/relate/blob/rfc-dbms-plugins/documentation/RFCs/dbms-plugins/dbms-plugins.rfc.md) for latest version 

### What is the new behavior?
(if this is a feature change)


documentation/RFCs/dbms-plugins/dbms-plugins.rfc.md
```diff
@@ -0,0 +1,153 @@
...
+- @relate Issue: (leave this empty)
+
+# Summary
+Support DBMS plugin management in the `@relate` framework. This RFC details the process for discovering available DBMS plugins, which Neo4j versions they support, as well as any and all DBMS configuration changes (in `neo4j.conf`) they require on installation.
+
+# Basic example
+Using `@relate/cli` as an example.
+```
+# list available plugin version sources
+$ relate environment:list-plugin-sources
```
nglgzz, line 12: Wouldn't it make more sense to have this command under the dbms topic instead of environment? Or maybe a completely separate one.

documentation/RFCs/dbms-plugins/dbms-plugins.rfc.md
```diff
@@ -0,0 +1,153 @@
...
+$ relate environment:remove-plugin-source 
+
+# list installed plugins for a given DBMS
+$ relate dbms:list-plugins <dbmsid> 
+
+# install/upgrade plugin for a given DBMS
+$ relate dbms:install-plugin <dbmsid>
+
+# uninstall plugin for a given DBMS
+$ relate dbms:uninstall-plugin <dbmsid>
```
nglgzz, line 27: Since we only need the DBMS ID we might want to allow passing in a list of IDs if we want to install the same plugin in multiple DBMSs.

documentation/RFCs/dbms-plugins/dbms-plugins.rfc.md
```diff
@@ -0,0 +1,153 @@
...
+    neo4jVersion: string; // supported DBMS semver (specific or range)
+    homepage: string; // plugin homepage URL
+    url: string; // plugin download URL
+    sha256: string; // sha256b checksum
+    config?: {[key: string]: string | string[]} // neo4j.conf changes needed
+}
+```
+Most of the above is self explanatory, however there is some additional semantics/logic to highlight in the `config` key.
+
+The `config` key allows plugins to declare changes to `neo4j.conf` a plugin needs.
```
nglgzz, line 60: Really interesting approach! +1 for key operators.

documentation/RFCs/dbms-plugins/dbms-plugins.rfc.md
```diff
@@ -0,0 +1,153 @@
...
+    type: EXTENSION_TYPES;
+    main: string;
+    // new values:
+    dependencies?: {
+        plugins?: string[] // DBMS plugins
+        extensions?: : string[] // @relate extensions
+    }
+}
+```
+It's unclear if these dependencies also need to specify a minimum version of a plugin/extension. With this information when an extension (eg. graph-app) installs we can also add any missing dependencies.
```
nglgzz, line 116: I would expect the possibility to both pin a version or specify a minimum version. If the dependency introduces breaking changes it shouldn't affect the extension unless they "upgrade" that dependency.

documentation/RFCs/dbms-plugins/dbms-plugins.rfc.md
```diff
@@ -0,0 +1,153 @@
...
+    - `string`: Overwrite any existing values
+    - `string[]`: Add any non-existing values (comma separated)
+- `-:` (remove)
+    - `boolean`: Remove if key present
+    - `string`: Remove if matches existing value
+    - `string[]`: Remove matching values (comma separated)
+
+Finally we also recognize that users are not always able/willing to make changes to the config, thus we need a way to request changes (i.e. make them optional but recommended):
+- `+?:` request to add key
+- `-?:` request to remove key
```
oskarhane, line 94: This looks like the syntax used in TC 👍, is this an established format?

documentation/RFCs/dbms-plugins/dbms-plugins.rfc.md
oskarhane: This could get really hairy. 

I don't think a plugin should have multiple versions that supports the same version of Neo4j.
So in the plugin case, I don't think it makes sense with a minimum version.

In the relate extension case, possibly, and makes me wonder if we should focus on DBMS plugins in this RFC and look into dependencies on extensions later.

oskarhane (commented):
This is very good!
I like the config change syntax, that should cover all cases.

I'm not sure about is we want to bundle plugin and extension dependencies in the same RFC as they behave a bit differently (plugins tied to DBMS version, relate extensions not) and I don't want to block adding plugin support for figuring out how to handle extension dependencies.

documentation/RFCs/dbms-plugins/dbms-plugins.rfc.md
jk05: +1

documentation/RFCs/dbms-plugins/dbms-plugins.rfc.md
```diff
@@ -0,0 +1,153 @@
...
+We propose the creation of an official Neo4j plugin indirection file that lists officially supported plugins and where to find a comprehensive list of their versions. This file is modelled after the [current Neo4jLabs plugins.json](https://github.com/neo4j/docker-neo4j/blob/663161f5a7847bff9dcdd1e164219ba3558663bf/neo4jlabs-plugins.json).
+
+```Typescript
+interface IDBMSPluginIndirection {
+    name: string; // plugin name
+    homepage: string; // plugin homepage, users are directed here for information/support
+    versions: string; // URL from which to fetch available plugin versions
+}
+```
+Relate will consume the indirection file over HTTP out of the box, and offer users the ability to add/remove additional sources as needed.
```
jk05, line 44: remove officially supported ones too?

documentation/RFCs/dbms-plugins/dbms-plugins.rfc.md
jk05: +1.  Not sure how we do this currently, this seems like a way of fine tuning the config changes needed. So would these have to be written/updated by plugin authors?

documentation/RFCs/dbms-plugins/dbms-plugins.rfc.md
huboneo: I thought about it. The reason I added these to environment is that I can see a case where we want to have different plugin sources per env. WDYT?

documentation/RFCs/dbms-plugins/dbms-plugins.rfc.md
huboneo: I'm not sure that's true, Should we maybe also have the name of the plugin?

documentation/RFCs/dbms-plugins/dbms-plugins.rfc.md
huboneo: @oskarhane I agree that we can defer relate extensions to another RFC, I'll update!

documentation/RFCs/dbms-plugins/dbms-plugins.rfc.md
huboneo: Thats where I took it from, not sure if it's just a TC thing or something else

documentation/RFCs/dbms-plugins/dbms-plugins.rfc.md
```diff
@@ -0,0 +1,135 @@
...
+Finally we also recognize that users are not always able/willing to make changes to the config, thus we need a way to request changes (i.e. make them optional but recommended):
+- `+?:` request to add key
+- `-?:` request to remove key
+
+If a plugin does not need config changes, the config property can safely be omitted from the version schema.
+
+It's unclear if dependencies also need to specify a minimum version of a plugin. With this information when an extension (eg. graph-app) installs we can also add any missing DBMS plugins. 
+
+Today there is no direct link between relate extensions and DBMSs. This means that when for example Bloom installs it is added on a system level and not for a specific DBMS.
+
```
huboneo, line 101: These two points above are unclear to me, anyone has a good idea?

documentation/RFCs/dbms-plugins/dbms-plugins.rfc.md
nglgzz: Right, I missed that. I still think we could have a list of DBMSs though. Something like: 

```
relate dbms:install-plugin <plugin name> <dbmsids>
```

That way if someone wants, they can easily install/uninstall the same plugin to many (or all) DBMSs.

documentation/RFCs/dbms-plugins/dbms-plugins.rfc.md
jk05: ignore this. I missed the part about additional sources

documentation/RFCs/dbms-plugins/dbms-plugins.rfc.md
nglgzz: 1. Wouldn't the version of the plugin we install depend on the version of the DBMS? Making it unnecessary for extensions to specify a minimum version?

2. I don't think we need an explicit link between extensions and DBMSs. Tags can be used for this. So in your example Bloom would add a tag to recognize the DBMSs it installed.

documentation/RFCs/dbms-plugins/dbms-plugins.rfc.md
nglgzz: FWIW, I think graph apps and extensions should be different entities. 

Graph apps are hosted files and can be discovered during runtime, while extensions are NestJS modules that need to be known before the application is started. For the same reason, I think extensions (and not graph-apps) should be global and not depending on a specific environment. It would remove the awkward `defaultEnvironment` config we currently have, and might also make it easier to think about plugins.

documentation/RFCs/dbms-plugins/dbms-plugins.rfc.md
oskarhane: +1 what @nglgzz said.
Bloom users should know that Bloom works with Neo4j 4.1 (just making this up) and not 3.5. Plugin version is irrelevant, there's only one plugin version per Neo4j version.

I don't understand the issue that

> Today there is no direct link between relate extensions and DBMSs. This means that when for example Bloom installs it is added on a system level and not for a specific DBMS.

brings.
Plugin dependencies for graph-apps should be automatically installed to DBMS:s when the user wants to use the graph-app for a specific dbms, not at graph-app installation, right?

documentation/RFCs/dbms-plugins/dbms-plugins.rfc.md
huboneo: @nglgzz that sounds resonable. The only reason they are the same today is technical, I see no reason moving them elsewhere. What is your suggestion?

documentation/RFCs/dbms-plugins/dbms-plugins.rfc.md
huboneo: @oskarhane this sentence refers to the case where a user does not use `app:open` (or other relate path) but instead tries to access a DBMS using raw credentials for example. Just a case I thought of, not sure it's relevant

huboneo: @nglgzz not sure tags are needed, we should be able to list installed plugins for a given DBMS?

nglgzz: @huboneo Yeah never mind that part, I initially misunderstood your second point. 

I think installing plugin dependencies only when launching apps makes sense. And additionally, we could have a command/method for installing app dependencies for specific DBMSs.

nglgzz (approved):
Looks good to me. We can discuss the open questions about extensions/apps with a separate RFC.

PR #247, opened 2020-12-22
Reduce manifest reads

nglgzz:

Performance improvements

- When initializing an environment, all DBMS manifests in that environment are read.
- When linking a DBMS, all DBMS manifests are read.
- When getting a DBMS, all DBMS manifests are read.
- `resolveProjects` reads all project manifests to retrieve one project.
- No guards against multiple projects having the same name.

### What is the new behavior?
- Initializing an environment doesn't trigger any DBMS reads.
- Linking a DBMS gets a DBMS instead of listing all of them.
- When getting a DBMS, if the DBMS exists in memory, no manifests are read. Otherwise, all DBMS manifests are read.
- `resolveProjects` works as `resolveDbms` without reading any manifest.
- Projects behavior is now consistent with DBMSs in case of duplicate names. 

No

### Other information:
Further optimizations can be done, especially for projects, but they would require some more refactoring, so leaving that for later, 
considering this should be a good start in reducing the chance of reading and writing to a manifest at the same time.
 
I measured the impact of these changes by running a few CLI commands and using the `MANIFEST_READ` hook to count the number of total reads. And then compared the results between this branch and master. I tested with one environment, 3 projects, and 6 DBMSs. Below are the results:

| Command         | # of reads (master) | # of reads (this PR) |
|-----------------|---------------------|----------------------|
| dbms:info          | 24                  | 6                    |
| dbms:start [id] | 18                  | 6                    |
| dbms:stop [id]  | 18                  | 6                    |
| dbms:start         | 24                  | 6                    |
| dbms:stop         | 24                  | 6                    |
| project:list          | 21                  | 3                    |

Note that the CLI commands are spawning an entire application for each command, so the difference in performance is not as dramatic as it would be in a long-running application (eg. Neo4j Desktop).

### DIFF
--- a/packages/common/src/entities/dbmss/dbmss.local.ts
+++ b/packages/common/src/entities/dbmss/dbmss.local.ts
@@ -438,7 +438,7 @@ export class LocalDbmss extends DbmssAbstract<LocalEnvironment> {
                 }
 
                 return {
-                    connectionUri: dbms.connectionUri,
+                    connectionUri: await this.getConnectionUri(dbms.id),
                     description: dbms.description,
                     edition: v?.edition,
                     id: dbms.id,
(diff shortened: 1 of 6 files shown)


nglgzz:
[Canary 1.3.12-dev4 remote connection to aura db failed](https://trello.com/c/aJLKDv5s/1544-canary-1312-dev4-remote-connection-to-aura-db-failed)

linuslundahl (approved):
Nice work on this! Let's get this in and then continue with the "Replace dbmss.get with this method for further performance gains." when we have more time.

nglgzz:
[Address concurrent manifest writes and reads](https://trello.com/c/0CsEUTA4/205-address-concurrent-manifest-writes-and-reads)

PR #250, opened 2021-01-14
CLI: Plugins plugin

akollegger:
- [ ] The commit message follows our [guidelines](https://github.com/neo4j-devtools/relate/blob/master/CONTRIBUTING.md)

feature(cli): add `@oclif/plugin-plugins` to enable cli plugin commands for easy extension of the CLI. 

No easily managed user interaction. 

### What is the new behavior?

```
> ./bin/run plugins --help
(relate)-[plugins]->(plugin)

list installed plugins

USAGE
  $ relate plugins

OPTIONS
  --core  show core plugins

EXAMPLE
  $ relate plugins

COMMANDS
  plugins:install    installs a plugin into the CLI
  plugins:link       links a plugin into the CLI for development
  plugins:uninstall  removes a plugin from the CLI
  plugins:update     update installed plugins
```

No, this addition does not change existing behaviors.


huboneo:
Not sure "plugin" is a good nomenclature here?

nglgzz:
Yeah, plugins are DBMS plugins. Since we have multiple points of extension for Relate and the DBMS we need to be very clear about their difference. I don't have many alternatives in mind, but `gcloud` calls them components, so maybe we can follow that same naming?

akollegger:
I agree that "plugins" is a problematic term. This is a simplistic integration with oclif's plugin framework. I'll take a look at aliasing the commands.

FWIW I like @nglgzz 's suggestion of `components`, which users probably should expect to be any of the relate extensions (not just cli extensions).

PR #254, opened 2021-01-26
Fix downloading of sample projects.

linuslundahl:
This addresses an issue where sample projects couldn't be downloaded when the default branch had been renamed from master to something else.

This is a fix for installing Sample projects. It changes the assumption that the default branch of a repo at `neo4j-graph-examples` is `master` and now instead uses `default_branch` from the GitHub API to determine the download URL.

`master` is hardcoded in the download URL.

### What is the new behavior?
(if this is a feature change)


nglgzz (approved):
Looks good to me

PR #258, opened 2021-02-10
Optimize listing of project files

nglgzz:

Performance improvement

- Adding a big directory (eg. the relate repository containing over 200k files) as a project and then listing the files in that project makes the process run out of memory and crash. Filtering out the directories containing most files (ie. node_modules) has no effect on the outcome.
- If there is a directory the user has no read access to inside the project, listing files will fail.

### What is the new behavior?
- The process no longer crashes with big directories, and when filtering out directories, those directories are not read at all.
- If the user has no access to a directory inside their project, that directory is ignored.

No


packages/cli/src/modules/project/list-files.module.ts
```diff
@@ -19,13 +19,43 @@ export class ListFilesModule implements OnApplicationBootstrap {
...
         const environment = await this.systemProvider.getEnvironment(environmentId);
         const projectId = flags.project || (await selectProjectPrompt('Select a Project', environment));
-        const files = await environment.projects.listFiles(projectId);
+
+        let {ignore = []} = flags;
+        // This is to accept both common formats for arrays in the CLI:
+        // command -i val1 -i val2 -i val3
+        // command -i val1,val2,val3
+        if (ignore.length === 1) {
+            ignore = ignore[0].split(',');
```
jk05, line 31: 👍

packages/cli/src/modules/project/list-files.module.ts
```diff
@@ -19,13 +19,43 @@ export class ListFilesModule implements OnApplicationBootstrap {
...
+                      field: 'directory',
+                      type: FILTER_COMPARATORS.NOT_CONTAINS,
+                      value: 'node_modules',
+                  },
+                  {
+                      field: 'directory',
+                      type: FILTER_COMPARATORS.NOT_CONTAINS,
+                      value: '.git',
+                  },
+              ];
```
jk05, line 52: should we do this within Relate itself?

packages/cli/docs/project.md
```diff
@@ -178,7 +178,16 @@ USAGE
...
+  -x, --extended                 show extra columns
+  --columns=columns              only show provided columns (comma-separated)
+  --filter=filter                filter property by partial string matching, ex: name=foo
+  --no-header                    hide table header from output
+  --no-truncate                  do not truncate output to fit screen
+  --output=csv|json|yaml         output in a more machine friendly format
+  --sort=sort                    property to sort by (prepend '-' for descending)
 
 EXAMPLES
   $ relate project:list-files
```
jk05, line 193: add an example in particular for `ignore`?

packages/cli/src/modules/project/list-files.module.ts
nglgzz: I don't think we should add this to the common package, but it feels like a sensible default for the CLI. I will change this to be a default value for the flag though, in the current format there's no way for the user to know which folders we're ignoring by default.

jk05: right in terms of the CLI output, Im with you.

jk05 (approved):

PR #259, opened 2021-02-15
Add plugin sources list

nglgzz:

Adds a file containing the official plugin sources for Relate.

-

### What is the new behavior?
-

No

### Other information:
-

### DIFF
--- a/plugins.json
+++ b/plugins.json
@@ -1,0 +1,18 @@
+{
+    "apoc": {
+        "homepageUrl": "https://github.com/neo4j-contrib/neo4j-apoc-procedures",
+        "versionsUrl": "https://neo4j-contrib.github.io/neo4j-apoc-procedures/versions.json"
+    },
+    "streams": {
+        "homepageUrl": "https://github.com/neo4j-contrib/neo4j-streams/",
+        "versionsUrl": "https://neo4j-contrib.github.io/neo4j-streams/versions.json"
+    },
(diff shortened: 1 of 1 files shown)


huboneo:
Why a file and not an endpoint?

nglgzz:
Where would we have the endpoint?

I was thinking to use a file because:
- We can just get it from https://raw.githubusercontent.com/neo4j-devtools/relate/master/plugins.json once this is merged.
- It's simple enough to update, just create a PR and when it's merged all versions of relate will be able to pull the updated plugin sources.
- This is also how it's currently implemented for both the plugin sources file that the neo4j docker image is using and versions files that plugins host.

Is there anything I'm missing?

huboneo:
yeah sure makes sense, just check github terms on using their repos for prod purposes. Any rate limits?

nglgzz:
Maybe I should have also added in the description that another PR will follow this, to change the `NEO4J_PLUGIN_SOURCES_URL` constant to point to this file and to remove the mocks we currently have in the e2e tests for plugin sources.

nglgzz:
Good point, will check that

jk05 (approved):

nglgzz:
I couldn't find anything official specific to the raw.githubusercontent.com endpoint, and the little information I found was conflicting, so it's probably best to create an endpoint like you suggested @huboneo.

PR #264, opened 2021-02-24
Upgrade plugins when upgrading a DBMS

nglgzz:

Feature

When upgrading a DBMS, plugins are copied as-is, even though they might be incompatible with the version the DBMS is being upgraded to.

When uninstalling a running DBMS on Windows, the status check would be performed incorrectly so the DBMS would always appear as stopped and the uninstall would proceed until a filesystem error is thrown because the neo4j.conf file would be in use.

### What is the new behavior?
A new method is available to get a preview of the plugin upgrades that would be performed during a DBMS upgrade. 

Users have a choice when upgrading a DBMS to:
- Discard all plugins.
- Upgrade all plugins - plugins will be upgraded to a compatible version when one is found and copied as-is when one isn't found.
- Upgrade only upgradable plugins - plugins will be upgraded to a compatible version when one is found, and discarded otherwise.

In all cases, all plugins are copied to a folder the DBMS doesn't know about, so the user can still access the plugins that were available before the upgrade.

When uninstalling a running DBMS on Windows, the status check works as intended, and the same error that is thrown in other platforms is thrown. 

Yes, the `dbmss.upgrade` method now has an options object instead of multiple boolean arguments.

### Other information:
As part of this PR I also extracted the DBMS upgrade logic to a util file, this is because the `dbmss.local` file is getting huge and difficult to navigate. I did this in the first commit, and added changes to the method after that, so it might be easier to see what changed by looking at the changes for the first commit first, and then all the rest.


packages/common/src/entities/environments/environment.constants.ts
```diff
@@ -50,6 +50,7 @@ export const NEO4J_IMPORT_DIR = 'import';
 export const NEO4J_RUN_DIR = 'run';
 export const NEO4J_LIB_DIR = 'lib';
 export const NEO4J_PLUGIN_DIR = 'plugins';
+export const NEO4J_PLUGIN_AVAILABLE_DIR = 'plugins-available';
```
nglgzz, line 53: Not entirely sure about this name, suggestions welcome

packages/common/src/models/dbms.model.ts
```diff
@@ -27,6 +27,19 @@ export interface IDbmsVersion {
...
+    ALL = 'ALL',
+    NONE = 'NONE',
+    UPGRADABLE = 'UPGRADABLE',
+}
+
+export interface IDbmsUpgradeOptions {
+    noCache?: boolean;
+    migrate?: boolean;
+    backup?: boolean;
+    pluginUpgradeMode?: PLUGIN_UPGRADE_MODE;
```
nglgzz, line 40: Same for this, not convinced by this name

packages/common/src/entities/dbms-plugins/dbms-plugins.local.ts
```diff
@@ -165,7 +192,12 @@ export class LocalDbmsPlugins extends DbmsPluginsAbstract<LocalEnvironment> {
                     throw new NotFoundError(`Could not find DBMS root path for "${dbms.name}"`);
                 }
 
-                const pluginToInstall = getLatestCompatibleVersion(dbms, pluginSource, pluginVersions);
+                const dbmsVersion = dbms.version;
+                if (!dbmsVersion) {
```
jk05, line 196: what could have happened to get to this situation? How to resolve for a user?

packages/common/src/entities/environments/environment.constants.ts
jk05: what is this used for? wasn't obvious to me from the code

packages/common/src/models/dbms.model.ts
jk05: I think the name here is ok. When it comes to usage from the CLI or Desktop, we need to explain what  `ALL`, `NONE` and `UPGRADABLE` mean as some of those are not immediately obvious.  I guess `previewUpgrade` can help with that

packages/common/src/utils/dbmss/upgrade-neo4j.ts
```diff
@@ -0,0 +1,151 @@
...
+        return;
+    }
+
+    const existingPlugins = await env.dbmsPlugins.list(originalDbms.id);
+    await existingPlugins
+        .mapEach(async (plugin) => {
+            try {
+                await env.dbmsPlugins.install([upgradedDbms.id], plugin.name);
+            } catch (err) {
+                await emitHookEvent(HOOK_EVENTS.DEBUG, `could not install plugin ${plugin.name}: ${err}`);
```
jk05, line 30: so in this case for `PLUGIN_UPGRADE_MODE.UPGRADABLE`, we wouldn't do much more since I guess we can't?

nglgzz: Yeah, that's the idea. `PLUGIN_UPGRADE_MODE.UPGRADABLE` will only upgrade plugins that have compatible versions available and ignore all the rest.

packages/common/src/entities/dbms-plugins/dbms-plugins.local.test.ts
```diff
@@ -210,6 +210,28 @@ describe('LocalDbmsPlugins', () => {
...
+        expect(upgradableMapped).toEqual([
+            {
+                name: 'apoc',
+                installed: '4.0.0.17',
+                upgradable: '4.2.0.0',
+            },
+            {
+                name: 'neo4j-jwt-addon',
+                installed: '1.0.1',
+                upgradable: undefined,
```
jk05, line 230: Just a comment - I see we get `undefined` here in this case due to the optional chaining for this test which is fine. In CLI usage for example, we would to keep it `null` or something similar to make it obvious.

packages/common/src/utils/dbmss/upgrade-neo4j.ts
jk05: right yeah that makes sense

packages/common/src/entities/dbms-plugins/dbms-plugins.local.ts
nglgzz: To get here we would have had an error getting the version, root path, or edition. The user would just retry the operation. This is just the type that is returned by `dbmss.info`. 

I think it could be improved, but it would require making `dbmss.list` use `dbmss.get` instead of `dbmss.info`, and we'd need to add version and edition to the IDbms interface.

jk05: thats fine, I think leave this for now. I was just wondering if it could point towards a larger issue.

packages/common/src/entities/environments/environment.constants.ts
nglgzz: It's where we copy the plugins from the older version of the DBMS. They might or might not be compatible with the upgraded version, and they might be installed by Relate or manualy. We're just saving them here on upgrades so we're not deleting user data in case some of these were added manually.

packages/common/src/entities/dbms-plugins/dbms-plugins.local.test.ts
nglgzz: Good point :+1:

packages/common/src/entities/environments/environment.constants.ts
jk05: so are we planning to do something with this directory after? For restoring? Or just let users know thats where they are?

nglgzz: I think it's just for users. We'd let them know that their plugins are there in case there are any issues with the plugin upgrade.

jk05 (dismissed):

nglgzz:
[[Kick-off] relate handling DBMS plugins](https://trello.com/c/LDjRztj5/1569-kick-off-relate-handling-dbms-plugins)

jk05 (approved):

PR #265, opened 2021-03-01
CLI: simplify environment initialization

nglgzz:

API improvement

In order to get started with Relate, new users need to: 

- Run the `env:init` command, reply to prompts about configuration options that might not be relevant to them, or that they might not understand.
- Run the `env:use` command, to set the newly created environment as active.
- If the user doesn't have Desktop installed, they'll also have to either install Java on their machine or download it and extract it in Relate's cache.

### What is the new behavior?
In order to get started with Relate, new users need to: 

- Run `relate env:init <environmentName> --use`

Default values will be used for configuration and Java will be automatically downloaded in Relate's cache if it's not already present. Also with the `--use` flag, the environment just created will also be set as active. 

For advanced users the configuration prompts are still available with the `--interactive` or `-i` flag.

Yes, the arguments and flags for `relate env:init` have changed.

### DIFF
--- a/packages/cli/docs/environment.md
+++ b/packages/cli/docs/environment.md
@@ -51,6 +51,7 @@ ARGUMENTS
 
 OPTIONS
   -i, --interactive      Get prompted for each configuration option available
+  --noRuntime            Skip downloading the Java runtime required by the DBMS
   --type=(LOCAL|REMOTE)  Type of environment
   --use                  Set environment as active right after creating it
 
--- a/packages/cli/docs/environment.md
+++ b/packages/cli/docs/environment.md
@@ -4,7 +4,7 @@
 Managed sets of related resources and services, which may be local or remote.
 
 * [`relate environment:api-token [CLIENTID]`](#relate-environmentapi-token-clientid)
-* [`relate environment:init`](#relate-environmentinit)
+* [`relate environment:init [ENVIRONMENT] [HTTPORIGIN]`](#relate-environmentinit-environment-httporigin)
 * [`relate environment:list`](#relate-environmentlist)
 * [`relate environment:login [ENVIRONMENT]`](#relate-environmentlogin-environment)
 * [`relate environment:open [ENVIRONMENT]`](#relate-environmentopen-environment)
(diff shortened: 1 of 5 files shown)


packages/cli/docs/environment.md
```diff
@@ -37,26 +37,32 @@ EXAMPLES
...
 ALIASES
   $ relate env:init
 
 EXAMPLES
-  $ relate env:init
-  $ relate env:init --name=local-environment-name --type=LOCAL
-  $ relate env:init --name=remote-environment-name --type=REMOTE --httpOrigin=https://url.of.hosted.relate.com
+  $ relate env:init local-environment-name
+  $ relate env:init local-environment-name --use
+  $ relate env:init remote-environment-name https://url.of.hosted.relate.com --type=REMOTE
```
jk05, line 64: second arg would imply `--type=REMOTE` anyway?

packages/cli/src/modules/environment/environment.local.e2e.ts
```diff
@@ -24,11 +26,23 @@ describe('$relate environment', () => {
...
+                expect(ctx.stdout).not.toContain(`Environment "${ENV_NAME}" is now set as active.`);
+            });
+
+        test.stdout()
+            .stderr()
+            .it('creates environment and activates it', async (ctx) => {
+                await InitCommand.run([`${ENV_NAME_2}`, '--type=LOCAL', '--use']);
+
+                expect(ctx.stderr).toContain('done');
+                expect(ctx.stdout).toContain(`Environment "${ENV_NAME_2}" is now set as active.`);
```
jk05, line 44: 👍

packages/cli/docs/environment.md
nglgzz: That's true, though I'd prefer not to change the default values of flags implicitly. It's harder to explain to the user and it can lead to confusion.

We could get rid of the `--type` flag entirely, but then we're also committing to having only two types of environments. This is not necessarily bad I just don't know if we want to go this route.

jk05: I understand the point you're making. Can leave it as it is for now. Not sure what other types of environments there could be at this point anyway.

jk05 (approved):

PR #266, opened 2021-03-03
Update READMEs

nglgzz:

Docs update

The READMEs don't provide much useful information to new users.

### What is the new behavior?
- The main README describes what Relate is, the main surfaces we expose, and basic concepts. 
- CLI and web READMEs include a guide for getting started and a description of how to find API documentation.

No

### Other information:
The README for the common package still needs to be revisited, and it would probably make sense to add examples to show how it would work in different scenarios.

When reviewing this PR it might be easier to check the [file view](https://github.com/nglgzz/relate/blob/23968362c449e3535392ea898685c63fe6d4233c/README.md) rather than the diff.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,52 +1,66 @@
-# relate
-Just want to use it? See our [First use guide](./documentation/FIRST_USE_GUIDE.md) instead.
+# @relate
 
+A framework to quickly set up and manage Neo4j DBMS instances and related
+entities. Provides the same core features as Neo4j Desktop but through
+programmatic interfaces.
 
-### Installation
+Some examples of things that can be done with Relate:
 
-```
-$ npm ci
-$ npm link # makes the relate command available globally
-```
+-   Installing Neo4j DBMSs with their required Java runtime.
+-   Handle lifecycle operations (ie. start, stop, retrieving process and server status).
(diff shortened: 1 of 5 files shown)


oskarhane (approved):
Very nice!

README.md
```diff
@@ -1,52 +1,66 @@
...
+contain multiple databases.
 
-### Generate documentation
+### Project
 
-```
-$ npm run generate:docs
-```
+A project in Relate is a folder with some extra metadata and with possible
+associations to databases or DBMSs.
```
jk05, line 62: `files`?

jk05:
great stuff. Just an aside - the arguments in the CLI docs, e.g `ENTITYNAMEORID` or `BACKUPIDORPATH`, is it convention to write like this instead of snake case for example?

jk05 (approved):

nglgzz:
> great stuff. Just an aside - the arguments in the CLI docs, e.g `ENTITYNAMEORID` or `BACKUPIDORPATH`, is it convention to write like this instead of snake case for example?

No, you're right they should be in snake case. These are automatically generated by oclif from our camel case names, so I'm guessing that's why they don't have underscores, but I think there should be a way to override them.

PR #304, opened 2021-05-18
Handle JWT plugin as any other plugin

nglgzz:

Bug fix

Handling of the JWT plugin is hardcoded resulting in the same version being installed for all Neo4j versions. This breaks with 4.2 and 4.3 due to breaking changes that require a different version of the plugin.

### What is the new behavior?
The JWT plugin is handled like any other plugin, and the correct version of the plugin is installed depending on the Neo4j version. 

No


packages/common/src/entities/dbmss/dbmss.local.ts
```diff
@@ -525,68 +523,52 @@ export class LocalDbmss extends DbmssAbstract<LocalEnvironment> {
...
-        if (!authorizationProviders.includes(Str.from(securityPluginJavaPath))) {
-            neo4jConfig.set(
-                `dbms.security.authorization_providers`,
-                authorizationProviders
-                    .concat([securityPluginJavaPath])
-                    .join(',')
-                    .get(),
-            );
+
+        const authenticationProviders = neo4jConfig.get('dbms.security.authentication_providers') || '';
```
jk05, line 557: do you need the `|| ''` here?

packages/common/src/entities/dbms-plugins/dbms-plugins.local.test.ts
```diff
@@ -1,11 +1,12 @@
 import nock from 'nock';
+import fse from 'fs-extra';
 
 import {TargetExistsError} from '../../errors';
 import {IDbmsInfo, IDbmsPluginSource} from '../../models';
 import {waitForDbmsToBeOnline} from '../../utils/dbmss';
 import {dbReadQuery} from '../../utils/dbmss/system-db-query';
 import {TestEnvironment, TEST_NEO4J_CREDENTIALS} from '../../utils/system';
-import {NEO4J_PLUGIN_SOURCES_URL} from '../environments';
+import {NEO4J_JWT_ADDON_NAME, NEO4J_PLUGIN_SOURCES_URL} from '../environments';
```
jk05, line 9: whats the reason to call it `ADDON` vs `PLUGIN`?

jk05: SInce its being handled as a plugin, thats the only reason I asked. Its not a reason to change it, just leave it.

jk05 (approved):

PR #307, opened 2021-05-31
Store logs for failed upgrades

linuslundahl:
Adds a new directory in the relate data dir (`upgrade-logs`) that copies the logs from the DBMS failed to upgrade.

A new folder called `upload-logs` is created in the relate storage dir. Whenever an upgrade fails, the logs from the failed upgrade are copied into this directory to help users debug.

The failed upgrade neo4j directory is just deleted, removing all logs too. Leaving users without the logs for help when debugging.

### What is the new behavior?
Logs of failed upgrades are kept instead of deleted.

### DIFF
--- a/packages/cli/src/modules/db/dump.module.ts
+++ b/packages/cli/src/modules/db/dump.module.ts
@@ -41,7 +41,7 @@ export class DumpModule implements OnApplicationBootstrap {
 
         const dateISO = new Date().toISOString();
         const [date] = dateISO.split('.');
-        const file = path.join(to || `${dbms.name}-${database}-${date.replace(/:/g, '')}.dump`);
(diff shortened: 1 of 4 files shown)


nglgzz (commented):
This is a great change and it's gonna help a lot debugging failed upgrades!

What do you think about creating an additional `relate.debug.json` file in these folders with some extra debug information? I'm thinking the following could be useful:

- OS
- ID of the original DBMS
- source and target version
- upgrade options
- If there is a backup ID (or path) of the backup
- possibly the list of plugins in the original DBMS

linuslundahl:
> What do you think about creating an additional `relate.debug.json` file in these folders with some extra debug information? I'm thinking the following could be useful:

That's a great idea! I'll have a look at it.

nglgzz (approved):
:sparkles:

nglgzz (approved):

PR #315, opened 2021-06-22
Update Neo4j on Windows docs, add DBMS plugins docs

nglgzz:

Docs update

No

### DIFF
--- a/documentation/DBMS_PLUGINS.md
+++ b/documentation/DBMS_PLUGINS.md
@@ -51,9 +51,8 @@ here](https://dist.neo4j.org/relate/official-plugin-sources.json)) it's possible
 to add external plugin sources by pointing Relate to a path or URL containing a
 plugin source.
 
-Here is an example of a plugin source, a file containing the following, or a URL
-returning this object, would be a valid source that would allow Relate to manage
-the JWT plugin.
+**Example**: a file containing the following, or a URL returning this object, would be a
+valid source that would allow Relate to manage the JWT plugin.
 
 ```json
 {
--- a/documentation/DBMS_PLUGINS.md
+++ b/documentation/DBMS_PLUGINS.md
@@ -1,0 +1,139 @@
+# DBMS Plugins
+
+Extending DBMS functionality using plugins is common practice in the Neo4j
+community. @relate supports the discovery and management of DBMS plugins, to
+help users setup and consume Neo4j DBMSs, tools, and apps.
+
+## Commands
+
+Using the @relate/cli as an example:
+
+```bash
+# list available plugin sources
+$ relate dbms-plugin:list-sources
+
+# add plugin source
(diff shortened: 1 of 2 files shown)


documentation/NEO4J_ON_WINDOWS.md
```diff
@@ -104,3 +104,9 @@ this issue.
 Shell scripts on Windows are not executable on their own, they need to be
 attached to a shell. The shell window can be hidden, which is what we do, but it
 will still show for a fraction of a second.
+
+### Update June 2021
+
+Some versions of Neo4j (+4.1.10, +4.2.8, +4.3.0, enterprise only) now include
+the `unsupported.dbms.shutdown` procedure that can be used to gracefully
```
linuslundahl, line 111: Are we allowed to document this?

nglgzz: My understanding is that this shouldn't be documented in the DBMS since it's not supported and they can remove it at any point. I think we should document it here, so a year from now it won't take a huge effort to figure out how this works (since it's not so straightforward). 

Might be worth asking though.

documentation/DBMS_PLUGINS.md
```diff
@@ -0,0 +1,139 @@
...
+```
+
+In addition to the list of officially supported plugins (which can be [found
+here](https://dist.neo4j.org/relate/official-plugin-sources.json)) it's possible
+to add external plugin sources by pointing Relate to a path or URL containing a
+plugin source.
+
+Here is an example of a plugin source, a file containing the following, or a URL
+returning this object, would be a valid source that would allow Relate to manage
+the JWT plugin.
```
jk05, line 56: maybe needs rewording - `A file containing the following, or a URL
returning this object, would be an example of a valid source that would allow Relate to manage
the JWT plugin.` or something. And perhaps put it under an example heading? Not sure

emmaagh: Yeah, I think making this a little clearer would be nice. Could do

>For example, a file containing the following object, or a URL returning it, would be a valid source that would allow Relate to manage the JWT plugin.

nglgzz: Good point, updated

documentation/NEO4J_ON_WINDOWS.md
jk05: I think its worth asking about this as we don't want to get a barrage of messages at some point asking why its there.

nglgzz: Just asked and the intention is to not encourage people to use this procedure as it's internal, unsupported and subject to change. I added a note about that to make it more clear.

jk05 (approved):
🦾  😃  lgtm

emmaagh (approved):

PR #316, opened 2021-06-24
Replace hasha with crypto for sha256 check

nglgzz:

Bug fix

Recent versions of hasta don't play well with Electron, so when verifying a shasum (after downloading a plugin/DBMS/extension) on a packaged Electron app the method generating the hash hangs indefinitely.

### What is the new behavior?
To generate the hash of downloaded files we now use the core crypto module instead of hasha. 

No

### DIFF
--- a/packages/common/package-lock.json
+++ b/packages/common/package-lock.json
@@ -6,7 +6,7 @@
     "packages": {
         "": {
             "name": "@relate/common",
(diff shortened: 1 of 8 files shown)


emmaagh (approved):
👍

jk05 (approved):
Checked by linking in Desktop and creating a packaged app. Seems to work 👍

PR #318, opened 2021-07-23
Sign PowerShell script to start DBMSs

nglgzz:

Bug fix

The PowerShell script we use to start DBMSs is not signed, so in certain environments it fails to execute.

### What is the new behavior?
The PowerShell script is now signed.

No.

### Other information:
If we do further changes to this script in the future we'll have to sign it again. For now I did it manually, but if we start making more changes in the future (which I doubt will be the case) it might be worth adding this step as part of the release pipeline.

To test that the signature is valid run the following command in PowerShell:
```ps1
Get-AuthenticodeSignature .\packages\common\neo4j-start.ps1
```


emmaagh (approved):
🖋️

PR #319, opened 2021-07-23
Always copy DBMS starter PowerShell script when starting DBMSs

nglgzz:

Bug fix

The DBMS starter script is copied only when it doesn't already exist, this means that any changes we make to it will not be applied on upgrades of Relate, unless the script in the cache is manually deleted. 

### What is the new behavior?
The DBMS starter script is copied every time a DBMS is started on Windows. This way we ensure it's always up to date. 

No


emmaagh (approved):
Seems reasonable 👍

PR #331, opened 2021-09-08
dbms-plugins resolvers and tests added

jk05:

Added resolvers and tests for the following `dbms-plugin` operations:
- `add-sources`
- `list-sources`
- `remove-sources`
- `install`
- `list`
- `uninstall`

No resolvers or tests for and `dbms-plugins` operations 

### What is the new behavior?
Can now list/add/remove dbms-plugins and sources.

no

### DIFF
--- a/packages/web/src/entities/dbms-plugins/dbms-plugins.types.ts
+++ b/packages/web/src/entities/dbms-plugins/dbms-plugins.types.ts
@@ -1,0 +1,79 @@
+import {ArgsType, Field, InputType, ObjectType} from '@nestjs/graphql';
+import {DbmsPluginConfig, IDbmsPluginInstalled, IDbmsPluginSource, IDbmsPluginVersion} from '@relate/common';
+import {GraphQLJSONObject} from 'graphql-type-json';
+
+@ObjectType()
+export class DbmsPluginVersion implements IDbmsPluginVersion {
+    @Field(() => String)
+    version: string;
+
+    @Field(() => String)
+    neo4jVersion: string;
+
+    @Field(() => String, {nullable: true})
+    homepageUrl?: string;
+
+    @Field(() => String)
+    downloadUrl: string;
+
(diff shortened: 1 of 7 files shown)


packages/web/src/entities/dbms-plugins/dbms-plugins.types.ts
```diff
@@ -0,0 +1,79 @@
...
+
+    @Field(() => Boolean, {nullable: true})
+    isOfficial?: boolean;
+}
+
+@ObjectType()
+export class DbmsPluginSource extends DbmsPluginSourceBase {}
+
+@ObjectType()
+export class DbmsPluginInstalled extends DbmsPluginSource implements IDbmsPluginInstalled {
```
nglgzz, line 46: I understand they're pretty much the same, so it probably doesn't matter, but shouldn't this class extend `DbmsPluginSourceBase`?

packages/web/src/entities/dbms-plugins/dbms-plugins.types.ts
```diff
@@ -0,0 +1,79 @@
...
+    @Field(() => DbmsPluginVersion)
+    version: DbmsPluginVersion;
+}
+
+@ArgsType()
+export class InstallOrUninstallDbmsPluginArgs {
```
nglgzz, line 52: Since you only have the plugin name here and not all the args that are needed for installing/uninstalling plugins,  `PluginNameArgs` might be a more accurate name instead?

jk05: makes sense

packages/web/src/entities/dbms-plugins/dbms-plugins.types.ts
jk05: let me take a look, I cant recall why I did it this way

nglgzz (approved):

PR #336, opened 2021-10-01
Improve upgrade error logging

emmaagh:
### Please check if the PR fulfils these requirements

Feature

### What is the current behaviour?
If a DBMS folder is locked, the DBMS upgrade process fails midway through. An `EBUSY` error is thrown. The catch block tries to clean up but runs into problems:
- uninstalling the DBMS errors, and that error is swallowed
- moving the restored DBMS to the original folder errors `dest already exists` and that error is returned => so the original error is lost

No information on what happened is logged out.

### What is the new behaviour?
- The upgrade process begins by trying to open the DBMS folder. If that isn't possible, the upgrade is aborted and a clear error message is returned, explaining either that the file is locked (if the error is `EBUSY`), or cannot be opened
- The catch/cleanup block emits a DEBUG event so that we never completely lose the error even if the cleanup fails
- The uninstall DBMS lines in the catch/cleanup block which were swallowing errors now emit a DEBUG event
- The restore backup & move backup lines now also emit a DEBUG event: this isn't as important, as the error message is rethrown so the end user will see the error, but I thought it was useful for the logs to show the context

No

### Other information:
See https://github.com/neo-technology/neo4j-desktop/pull/1078 for matching desktop changes.

Ticket: https://trello.com/c/YvQxYvgU/1585-initiating-upgrade-with-windows-terminal-open-to-database-directory-causes-database-operations-to-fail

### DIFF
--- a/packages/common/src/utils/dbmss/upgrade-neo4j.ts
+++ b/packages/common/src/utils/dbmss/upgrade-neo4j.ts
@@ -56,6 +56,17 @@ export const upgradeNeo4j = async (
     version: string,
     options: IDbmsUpgradeOptions,
 ): Promise<IDbmsInfo> => {
+    const dbms = await env.dbmss.get(dbmsId);
+    const dbmsManifest = await env.dbmss.manifest.get(dbmsId);
+
+    if (semver.lte(version, dbms.version!)) {
+        throw new InvalidArgumentError(`Target version must be greater than ${dbms.version}`, ['Use valid version']);
+    }
+
+    if (dbms.status !== DBMS_STATUS.STOPPED) {
+        throw new InvalidArgumentError(`Can only upgrade stopped dbms`, ['Stop dbms']);
+    }
+
     const dbmsRootPath = env.dbmss.getDbmsRootPath(dbmsId);
     try {
         const fileDescriptor = await fse.open(dbmsRootPath, fse.constants.O_RDONLY | 0x10000000);
@@ -69,17 +80,6 @@ export const upgradeNeo4j = async (
                 : 'the DBMS folder could not be opened';
         const errorMessage = `Did not attempt to upgrade the DBMS because ${reason}. Folder location: ${dbmsRootPath}`;
         throw new DbmsUpgradeError(errorMessage, openError.message);
-    }
-
-    const dbms = await env.dbmss.get(dbmsId);
-    const dbmsManifest = await env.dbmss.manifest.get(dbmsId);
-
-    if (semver.lte(version, dbms.version!)) {
-        throw new InvalidArgumentError(`Target version must be greater than ${dbms.version}`, ['Use valid version']);
-    }
-
-    if (dbms.status !== DBMS_STATUS.STOPPED) {
-        throw new InvalidArgumentError(`Can only upgrade stopped dbms`, ['Stop dbms']);
     }
 
     const {entityType, entityId} = await emitHookEvent(HOOK_EVENTS.BACKUP_START, {
--- a/packages/common/src/utils/dbmss/upgrade-neo4j.ts
+++ b/packages/common/src/utils/dbmss/upgrade-neo4j.ts
@@ -50,6 +50,23 @@ const upgradePlugins = async (
         .unwindPromises();
 };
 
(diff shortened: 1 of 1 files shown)


packages/common/src/utils/dbmss/upgrade-neo4j.ts
```diff
@@ -56,6 +56,21 @@ export const upgradeNeo4j = async (
...
+    const dbmsRootPath = env.dbmss.getDbmsRootPath(dbmsId);
+    try {
+        const fileDescriptor = await fse.open(dbmsRootPath, fse.constants.O_RDONLY | 0x10000000);
+        await fse.close(fileDescriptor);
+    } catch (openError) {
+        const reason =
+            openError.code === 'EBUSY'
+                ? 'the DBMS folder is locked. This is probably because another program, ' +
+                  'for example your Terminal, is open and is using the folder. ' +
+                  'Close the program and retry the upgrade'
```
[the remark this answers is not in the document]
emmaagh: How about this as the error message:

>Did not attempt to upgrade the DBMS because the DBMS folder is open in another program. Close the program and retry the upgrade.

?

packages/common/src/utils/dbmss/upgrade-neo4j.ts
```diff
@@ -50,6 +50,23 @@ const upgradePlugins = async (
...
 };
 
+const checkDbmsFolderIsNotLocked = async (env: LocalEnvironment, dbmsId: string): Promise<void> => {
+    const dbmsRootPath = env.dbmss.getDbmsRootPath(dbmsId);
+    try {
+        const O_EXLOCK = 0x10000000;
+        // Attempt to open the folder to confirm it is accessible
+        const fileDescriptor = await fse.open(dbmsRootPath, fse.constants.O_RDONLY | O_EXLOCK);
+        await fse.close(fileDescriptor);
+    } catch (openError) {
```
jk05, line 60: does this have a type associated with it?  Or should we type as `unknown` and check instance of Error? Just putting out ideas

jk05: https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-0.html#unknown-on-catch-clause-bindings

emmaagh: Prettier won't let me set a type other than `any` or `unknown`. We have a similar thing in the next catch block (we do `e.message` without knowing the type of `e`) - since that was already there I just assumed we were OK with that potential issue haha. But yes, it would be nice to tidy it up. Checking instanceof `Error` isn't good enough, as `Error` doesn't have the `code` property. The actual error type is `NodeJS.ErrnoException`, but I'm having trouble accessing that (it should come with the `@types/node` module, and I've also seen advice about modifying `tsconfig`, but it's still not available). So it's being a bit fiddly but I'll keep looking into it.

jk05: No worries. It was a suggestion. I also meant checking for instance of some type of error object. Its not a blocker here since we have done it elsewhere but it did make me think that about the types in catch blocks.

emmaagh: >I also meant checking for instance of some type of error object

Yeah, that's what I'm trying, but I can't get hold of the error type which is being thrown :/ Just tried updating @types/node but that hasn't helped unfortunately. Might have to leave this one as it is.

packages/common/src/utils/dbmss/upgrade-neo4j.ts
```diff
@@ -129,14 +148,18 @@ export const upgradeNeo4j = async (
...
+        });
         if (e instanceof RelateBackupError) {
             throw e;
         }
 
         // Backup logs and add extra debug information if an upgrade fails
         if (upgradedDbmsInfo) {
-            const dbmsRootPath = env.dbmss.getDbmsRootPath(upgradedDbmsInfo.id);
-            if (dbmsRootPath) {
+            const upgradedDbmsRootPath = env.dbmss.getDbmsRootPath(upgradedDbmsInfo.id);
```
jk05, line 161: 👍🏽

nglgzz (dismissed):

nglgzz (approved):

PR #337, opened 2021-10-04
dbmss info subscription

jk05:

- Adds a GraphQL Subscription for the dbmss info command
- Added `graphql-subscriptions` and `chokidar` in order to set up a watcher looking for `neo4j.pid` files being created / unlinked to establish dbms start / stop and publish an event

- No GraphQL subscriptions

### What is the new behavior?
- Can subscribe to `watchInfoDbmss` to get `DbmsInfo` for a relevant DBMS started / stopped

No

### Other information:
- https://github.com/graphql/graphql-playground/issues/1021 - if testing on GraphQL playground, can use a CSS change to be able to scroll. Should be fixed in newer versions.
- Left comments originating from NestJS docs around suggestions which we can implement in a later dev phase, shouldn't be a blocker right now
- Main issue is handling unsubscribing - removing the watcher etc. There is some discussion https://github.com/nestjs/graphql/issues/186 and I tried to implement it but couldn't get it to work yet - would prefer to get some ideas on this before I/we look into this any further as Im not sure I love this solution. Should we leave this for now and come back to it in another dev phase? Or should we put time into it now?
- Haven't looked into testing GraphQL subscriptions yet. Will pick it up as part  of this work or in cool down.


packages/web/src/entities/dbms/dbms.resolver.ts
```diff
@@ -71,6 +79,46 @@ export class DBMSResolver {
...
+    @Subscription(() => [DbmsInfo], {
+        resolve: (payload: any, _variables: any, context: any) => {
+            return context.environment.dbmss.info([payload.dbmsId]);
+        },
+    })
+    async [PUBLIC_GRAPHQL_METHODS.WATCH_INFO_DBMSS](
+        @Context('environment') environment: Environment,
+    ): Promise<AsyncIterator<unknown, any, undefined>> {
+        // @todo: this isn't fleshed out yet and is simplistic
+        // @todo: how to detect or handle unsubsribe? how to handle multiple watchers for an env?
```
nglgzz, line 91: Might be worth looking into this: https://github.com/nestjs/graphql/issues/186#issuecomment-478336248

packages/web/src/entities/dbms/dbms.resolver.ts
```diff
@@ -71,6 +79,46 @@ export class DBMSResolver {
...
+        if (!environmentWatchers.get(environment.id)) {
+            // set watcher for environment
+            const setWatcher = (watchPath: string) => {
+                const watcher = chokidar.watch(watchPath, {
+                    persistent: true,
+                    ignoreInitial: true,
+                });
+
+                const eventHandler = (e: string) => {
+                    const dbmsIdCheck = e.match(/dbms-([^/]*)/);
```
nglgzz, line 101:
```suggestion
                    const dbmsIdCheck = e.match(/dbms-([^/]+)/);
```

packages/web/src/entities/dbms/dbms.resolver.ts
```diff
@@ -71,6 +79,46 @@ export class DBMSResolver {
...
+        },
+    })
+    async [PUBLIC_GRAPHQL_METHODS.WATCH_INFO_DBMSS](
+        @Context('environment') environment: Environment,
+    ): Promise<AsyncIterator<unknown, any, undefined>> {
+        // @todo: this isn't fleshed out yet and is simplistic
+        // @todo: how to detect or handle unsubsribe? how to handle multiple watchers for an env?
+        if (!environmentWatchers.get(environment.id)) {
+            // set watcher for environment
+            const setWatcher = (watchPath: string) => {
```
nglgzz, line 94: Could be worth extracting this into a util

packages/web/src/entities/dbms/dbms.resolver.ts
jk05: Yeah I looked into this. I couldn't get it to work yet but can take another look.

packages/web/src/entities/dbms/dbms.resolver.ts
jk05: Ill see. It might become more obvious when we add another subscription later. Ill look into it.

packages/web/src/entities/dbms/dbms.resolver.ts
jk05: Left a note in the PR description. Just wondering if I should look into this and get it working now? Or save for next phase?

packages/web/src/entities/dbms/dbms.resolver.ts
```diff
@@ -71,6 +79,34 @@ export class DBMSResolver {
...
+    @Subscription(() => [DbmsInfo], {
+        resolve: (payload: any, _variables: any, context: any) => {
+            return context.environment.dbmss.info([payload.dbmsId]);
+        },
+    })
+    async [PUBLIC_GRAPHQL_METHODS.WATCH_INFO_DBMSS](
+        @Context('environment') environment: Environment,
+    ): Promise<AsyncIterator<unknown, any, undefined>> {
+        if (!environmentWatchers.get(environment.id)) {
+            const watcher = setWatcher(path.join(environment.dirPaths.dbmssData, '**', 'run', '*.pid'));
```
jk05, line 91: will assign this path to a variable somewhere better at the end.

nglgzz (approved):
Looks good :sparkles:

PR #340, opened 2021-11-12
Modify watchInfoDbmss subscription to account for DBMS install/uninstall

jk05:

- adds debounce to the file event handler
- publish for DBMS install/uninstall events
- return structure is different, with a property for each event type (see screenshots)
- Note - currently can only return the DBMS ID when an uninstall occurs since that is the only info we would have from Chokidar and so cannot use `dbmss.info` after this point with this DBMS ID (since its deleted).

- DBMS install / uninstall will publish any new events for the subscription resolver

### What is the new behavior?
- New structure with DBMS install / uninstall published 

Yes - when introduced we would need to make adjustments wherever the `watchInfoDbmss` subscription has been used -  Desktop 2.0 and the Sample App.

### Other information
Screenshot 2021-11-12 at 10 37 03
Screenshot 2021-11-12 at 10 36 51

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -143,10 +143,10 @@
         "max-depth": "error",
         "max-len": ["error", {"code": 120, "comments": 120, "ignoreUrls": true}],
         "max-lines": ["warn", 500],
-        "max-lines-per-function": ["warn", 100],
+        "max-lines-per-function": ["off"],
         "max-nested-callbacks": ["error", 5],
         "max-params": ["warn", 4],
-        "max-statements": "warn",
+        "max-statements": ["off"],
         "max-statements-per-line": "error",
         "new-parens": "error",
         // This rule is conflicting with the new behavior from prettier
--- a/packages/common/src/entities/dbms-plugins/dbms-plugins.local.test.ts
+++ b/packages/common/src/entities/dbms-plugins/dbms-plugins.local.test.ts
@@ -29,7 +29,6 @@ const TEST_SOURCE_2: IDbmsPluginSource = {
     versionsUrl: 'https://neo4j-contrib.github.io/test-plugin-2/versions.json',
 };
 
-// eslint-disable-next-line max-statements, max-lines-per-function
 describe('LocalDbmsPlugins', () => {
     let app: TestEnvironment;
     let dbms: IDbmsInfo;
--- a/packages/common/src/entities/dbms-plugins/dbms-plugins.local.test.ts
+++ b/packages/common/src/entities/dbms-plugins/dbms-plugins.local.test.ts
@@ -29,6 +29,7 @@ const TEST_SOURCE_2: IDbmsPluginSource = {
     versionsUrl: 'https://neo4j-contrib.github.io/test-plugin-2/versions.json',
 };
 
+// eslint-disable-next-line max-statements, max-lines-per-function
 describe('LocalDbmsPlugins', () => {
     let app: TestEnvironment;
     let dbms: IDbmsInfo;
--- a/packages/web/src/utils/file-watcher.utils.ts
+++ b/packages/web/src/utils/file-watcher.utils.ts
@@ -1,25 +1,65 @@
 import chokidar, {FSWatcher} from 'chokidar';
 import {PubSub} from 'graphql-subscriptions';
+import {debounce} from 'lodash';
+import path from 'path';
 
 // @todo: should replace this in production: https://docs.nestjs.com/graphql/subscriptions#code-first
 export const pubSub = new PubSub();
 
-export const setWatcher = (watchPath: string): FSWatcher => {
+export enum DBMS_EVENT_TYPE {
+    STARTED = 'started',
+    STOPPED = 'stopped',
+    INSTALLED = 'installed',
+    UNINSTALLED = 'uninstalled',
+}
+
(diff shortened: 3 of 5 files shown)


packages/web/src/utils/file-watcher.utils.ts
```diff
@@ -1,25 +1,65 @@
...
-export const setWatcher = (watchPath: string): FSWatcher => {
+export enum DBMS_EVENT_TYPE {
+    STARTED = 'started',
+    STOPPED = 'stopped',
+    INSTALLED = 'installed',
+    UNINSTALLED = 'uninstalled',
+}
+
+// @todo: currently geared towards watching DBMS files/folders specifically
+export const setWatcher = (watchPath: string[]): FSWatcher => {
```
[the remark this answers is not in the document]
jk05: yeah will rename

packages/common/src/entities/dbms-plugins/dbms-plugins.local.test.ts
```diff
@@ -29,6 +29,7 @@ const TEST_SOURCE_2: IDbmsPluginSource = {
     versionsUrl: 'https://neo4j-contrib.github.io/test-plugin-2/versions.json',
 };
 
+// eslint-disable-next-line max-statements, max-lines-per-function
```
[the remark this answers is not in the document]
jk05: I will disable

packages/web/src/utils/file-watcher.utils.ts
```diff
@@ -1,25 +1,65 @@
...
     });
 
-    // determine dbmsId from chokidar event
-    const eventHandler = (e: string) => {
-        const dbmsIdCheck = e.match(/dbms-([^/]+)/);
-        const dbmsId = dbmsIdCheck ? dbmsIdCheck[1] : '';
-        pubSub.publish('infoDbmssUpdate', {dbmsId});
-    };
+    const eventHandler = (eventType: DBMS_EVENT_TYPE) =>
+        debounce(
```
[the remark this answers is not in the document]
jk05: This was just an avenue i was exploring. Will be removing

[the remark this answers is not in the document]
jk05: left it in. The raw events show that for adding a directory, it seems to fire two duplicate events (just with different flag values). Combined with trailing set to true, this seems to work with the `dbms.info` call in the resolver. But needs more checking across.

.eslintrc.json
```diff
@@ -143,10 +143,10 @@
         "max-depth": "error",
         "max-len": ["error", {"code": 120, "comments": 120, "ignoreUrls": true}],
         "max-lines": ["warn", 500],
-        "max-lines-per-function": ["warn", 100],
+        "max-lines-per-function": ["off"],
         "max-nested-callbacks": ["error", 5],
         "max-params": ["warn", 4],
-        "max-statements": "warn",
+        "max-statements": ["off"],
```
nglgzz, line 149: :+1:

packages/web/src/utils/file-watcher.utils.ts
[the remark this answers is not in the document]
nglgzz: Actually, never mind. Just re-read the code and the debounced function is not shared between the different events, so the case I mentioned shouldn't be an issue.

nglgzz (approved):

packages/web/src/utils/file-watcher.utils.ts
```diff
@@ -1,25 +1,66 @@
...
     };
 
+    const dirEventHandler = (eventType: DBMS_EVENT_TYPE.INSTALLED | DBMS_EVENT_TYPE.UNINSTALLED) =>
+        // debounce for now as multiple events for addDir (different flags)
+        debounce(
+            (e: string) => {
+                // determine dbmsId from chokidar event
+                // only interested in the top level dir
+                // so ignore any events with a trailing slash
+                const dbmsDirIdCheck = e.match(/dbms-([\w-]+[^/]$)/);
```
linuslundahl, line 45: Are these ID's determined from the dbms directory? Won't that be an issue for linked DBMSs that doesn't contain the DBMS ID in the directory?

linuslundahl: Or is it the symlinked path which I guess will contain the ID?

jk05: I presume it would be symlinked to the Relate data dir and should contain this - unless you had something else in mind? I was thinking thats we do with linked project dirs.

linuslundahl: Yeah, I was just concerned that if I linked `~/Documents/My DBMS` it wouldn't contain the ID. But when it's symlinked the symlink in the relate-data folder is named after the ID right?

jk05: yeah I just tried it myself and that seems to be the case as expected. ID comes from the relate-data folder (folder name and the manifest)

jk05: Screenshot 2021-11-17 at 10 47 06
>

linuslundahl: 👍

linuslundahl (approved):
Good stuff! Let's test it in Desktop!

PR #341, opened 2021-11-18
Allow passing path to dbms install.

linuslundahl:

- allow DBMS install / uninstall from custom path

- cannot install to a custom path, only Relate data dir

### What is the new behavior?
- can pass the custom install path dbmss.install, will symlink from the Relate DBMS data dir and can also uninstall
- adds a `isCustomPathInstallation` to DBMS manifests

Not that I can think of

If we do decide to go with this, then `isCustomPathInstallation` would also exist on all manifests even on V1 so something to be aware of.


packages/common/src/entities/dbmss/dbmss.local.ts
```diff
@@ -131,11 +132,22 @@ export class LocalDbmss extends DbmssAbstract<LocalEnvironment> {
...
+                throw new NotAllowedError(`Unable to create "${name}" in the location "${installPath}": ${err}`);
+            });
+        }
+
         // version as a file path.
         if ((await fse.pathExists(version)) && (await fse.stat(version)).isFile()) {
             const tmpPath = path.join(this.environment.dirPaths.tmp, uuidv4());
             const {extractedDistPath} = await extractNeo4j(version, tmpPath);
-            const dbms = await this.installNeo4j(name, this.getDbmsRootPath(), extractedDistPath, credentials);
+            const dbms = await this.installNeo4j(
```
nglgzz, line 145: There's another call to `this.installNeo4j` a few lines later that also needs to be updated (can't link it or leave a comment on that line since it's unchanged).

Also you need to link the installed DBMS when using the `installPath` or it won't show up when listing all DBMSs.

nglgzz (commented):
Good changes, just a couple of comments. Would also be good to have tests for this.

packages/common/src/entities/dbmss/dbmss.local.ts
```diff
@@ -131,11 +132,22 @@ export class LocalDbmss extends DbmssAbstract<LocalEnvironment> {
             throw new DbmsExistsError(`DBMS with name "${name}" already exists`);
         }
 
+        if (installPath) {
+            await fse.access(installPath, fse.constants.W_OK, (err) => {
+                throw new NotAllowedError(`Unable to create "${name}" in the location "${installPath}": ${err}`);
+            });
+        }
```
nglgzz, line 139: If you pass a callback argument the function will not return a promise. 

```suggestion
            await fse.access(installPath, fse.constants.W_OK).catch((err) => {
                throw new NotAllowedError(`Unable to create "${name}" in the location "${installPath}": ${err}`);
            });
        }
```

packages/common/src/entities/dbmss/dbmss.local.ts
linuslundahl: I just put up what I had so far for @jk05 to continue on. But good pointers.

packages/common/src/entities/dbmss/dbmss.local.ts
jk05: yeah Ive sorted it now. I was just pushing a change to make sure I could push to the correct branch

packages/common/src/entities/dbmss/dbmss.local.ts
```diff
@@ -475,6 +488,13 @@ export class LocalDbmss extends DbmssAbstract<LocalEnvironment> {
         await fse.copy(extractedDistPath, path.join(dbmssDir, dbmsIdFilename));
 
         try {
+            // link the DBMS if installing to a custom path
+            if (path.join(dbmssDir, dbmsIdFilename) !== this.getDbmsRootPath(dbmsId)) {
+                const linkedDbmsInfo = await this.link(path.join(dbmssDir, dbmsIdFilename), name);
+                dbmsId = linkedDbmsInfo.id;
+                await this.manifest.update(dbmsId, {isCustomPathInstallation: true});
+            }
```
jk05, line 496: seems a little messy here.

Also wonder if a util or some other way is needed to determine installing to a custom path.

packages/common/src/entities/dbmss/dbmss.local.ts
```diff
@@ -523,7 +549,22 @@ export class LocalDbmss extends DbmssAbstract<LocalEnvironment> {
...
+        await fse
+            .unlink(dbmsRootPath)
+            .then(() => {
+                // remove target if custom DBMS install path was used
+                if (dbms.isCustomPathInstallation && targetPath) {
+                    return fse.remove(targetPath);
+                }
+                return Promise.resolve();
+            })
+            .catch(() => fse.remove(dbmsRootPath));
```
jk05, line 567: suggestions please

packages/common/src/entities/dbmss/dbmss.local.ts
```diff
@@ -510,6 +530,12 @@ export class LocalDbmss extends DbmssAbstract<LocalEnvironment> {
             return installed;
         } catch (error) {
             await fse.remove(path.join(dbmssDir, dbmsIdFilename));
+            // if installing to a custom path, remove the linked DBMS dir
+            if (path.join(dbmssDir, dbmsIdFilename) !== this.getDbmsRootPath(dbmsId)) {
+                if (await fse.pathExists(this.getDbmsRootPath(dbmsId))) {
+                    await fse.remove(this.getDbmsRootPath(dbmsId));
+                }
+            }
```
jk05, line 538: same comment as above

packages/common/src/entities/dbmss/dbmss.local.ts
```diff
@@ -114,6 +114,7 @@ export class LocalDbmss extends DbmssAbstract<LocalEnvironment> {
         credentials = '',
         overrideCache = false,
         limited = false,
+        installPath?: string,
```
jk05, line 117: too many parameters passed in - we should switch to an object or set limits in future.

jk05:
Will add some tests after a review and general direction established

packages/common/src/entities/dbmss/dbmss.local.ts
nglgzz: I think you can simplify the condition as:
```
dbmssDir !== this.getDbmsRootPath()
```

And as far as I remember `dbmss.link` preserves the ID if there is one, so there shouldn't be the need to reassign it.

jk05: Good call on the simplification.

As for the ID, I didnt find that to be the case. I will check again, there may be something else going on I will double check

jk05: I think I know where the ID issue is coming from here. `manifest.update` (which contains `ensureFile`) uses `entityRootPath` so doesn't work with a custom path before linking. I will have a think about ways to deal with this. Besides I am placing it after the linking here on line 498, but need to do it before.

nglgzz: What's the problem with updating the manifest after linking?

jk05: Because no manifest exists prior to linking, and so a new id is created and used for the manifest.

jk05: Maybe theres another way to handle this but I am currently thinking create a simple manifest directly here using `fse.writeJson`, or adjust `manifest.update` to override with a custom path to install a manifest to, or a new method for manifests to simply create a manifest to a path?

packages/common/src/entities/dbmss/dbmss.local.ts
nglgzz: Need to check what happens on `fse.readlink` if the symlink is broken (eg. the target path was moved/deleted by the user), and if `fse.unlink` works to delete a broken link (since we're deleting the target path before the symlink), but I figure this could work: 

```suggestion
        // Remove DBMS directory from custom installation path
        if (dbms.isCustomPathInstallation) {
            const targetPath = await fse.readlink(dbmsRootPath);
            await fse.remove(targetPath);
        }

        // Remove DBMS directory or symlink from Relate data dir
        await fse.unlink(dbmsRootPath).catch(() => fse.remove(dbmsRootPath));
```

jk05: Ill probably do it this. I dont think we would even necessarily get to the `readLink` if symlink is broken given `!found` above this block. But I will double check

packages/web/src/entities/dbms/dbms.types.ts
```diff
@@ -24,6 +24,9 @@ export class Dbms implements Omit<IDbms, 'config'> {
 
     @Field(() => String)
     connectionUri: string;
+
+    @Field(() => String)
+    isCustomPathInstallation?: boolean;
```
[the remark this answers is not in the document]
jk05: 👍🏽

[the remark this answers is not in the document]
jk05: `same here`? was there another you mentioned?

packages/common/src/entities/dbmss/dbmss.local.ts
```diff
@@ -131,11 +132,22 @@ export class LocalDbmss extends DbmssAbstract<LocalEnvironment> {
             throw new DbmsExistsError(`DBMS with name "${name}" already exists`);
         }
 
+        if (installPath) {
+            await fse.access(installPath, fse.constants.W_OK).catch((err) => {
+                throw new NotAllowedError(`Unable to create "${name}" in the location "${installPath}": ${err}`);
```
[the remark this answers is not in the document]
jk05: 👍🏽

packages/common/src/entities/dbmss/dbmss.local.ts
```diff
@@ -510,6 +537,14 @@ export class LocalDbmss extends DbmssAbstract<LocalEnvironment> {
             return installed;
         } catch (error) {
             await fse.remove(path.join(dbmssDir, dbmsIdFilename));
+            // if installing to a custom path, remove the linked DBMS dir
+            if (targetDir) {
+                if (!targetDirAlreadyExists) {
+                    await fse.remove(targetDir);
+                } else if (!(error instanceof AmbiguousTargetError)) {
```
jk05, line 544: did this for the situation on line 494 - we have to be careful to not start clearing out user dirs inadvertently. Could make a new error type for this? Or some discussion around our thoughts here?

packages/common/src/entities/dbmss/dbmss.local.ts
```diff
@@ -523,6 +558,13 @@ export class LocalDbmss extends DbmssAbstract<LocalEnvironment> {
             throw new AmbiguousTargetError(`DBMS ${dbmsId} not found`);
         }
 
+        // Remove DBMS directory from custom installation path
+        if (dbms.isCustomPathInstallation) {
+            const targetPath = await fse.readlink(dbmsRootPath);
+            await fse.remove(targetPath);
```
jk05, line 564: I think in this case we should be ok to remove the dir?

packages/common/src/entities/dbmss/dbmss.local.ts
nglgzz: `AmbiguousTargetError` is meant to be thrown when we have multiple targets and cannot tell which one we're supposed to use. I think `TargetExistsError` is more suitable for this situation.

Either way, it's quite easy to miss what's going on here with all the nesting, if you move the symlink/target dir creation to a separate try/catch, you should be able to simplify this quite a bit.

packages/common/src/entities/dbmss/dbmss.local.ts
```diff
@@ -469,12 +478,30 @@ export class LocalDbmss extends DbmssAbstract<LocalEnvironment> {
...
-        await fse.copy(extractedDistPath, path.join(dbmssDir, dbmsIdFilename));
+        let targetDirAlreadyExists = false;
 
         try {
+            if (targetDir) {
+                targetDirAlreadyExists = await fse.pathExists(targetDir);
+                if (!targetDirAlreadyExists) {
+                    await fse.ensureDir(targetDir);
+                }
+                const files = (await fse.readdir(targetDir)).filter((file) => !file.startsWith('.'));
```
nglgzz, line 492: Hidden files are still files and we should fail in that case too because the directory is not empty.

```suggestion
                const files = await fse.readdir(targetDir);
```

jk05: Yeah I understand, I can remove this. It may get annoying with things like `.DS_Store` on Mac for example. At the same time, don't want to start listing all the files in the directory in the error message as there could be too many. So for now, makes sense to remove it.

packages/common/src/entities/dbmss/dbmss.local.ts
jk05: Not a fan of removing user dirs but for now will leave this. In future, can look at options being passed around what to do with these dirs.

packages/common/src/entities/dbmss/dbmss.local.ts
jk05: Ill push up what I have - not sure its simpler

packages/common/src/entities/dbmss/dbmss.local.ts
nglgzz: I think it's fine to delete this folder. At this point it's a DBMS folder, with just DBMS data in it, not a folder containing user data. And if the user wants to uninstall the DBMS it only makes sense that the folder is deleted.

nglgzz (approved):

packages/common/src/entities/dbmss/install.test.ts
```diff
@@ -129,6 +130,71 @@ describe('LocalDbmss - install', () => {
...
+            false,
+            false,
+            testTargetPath,
+        );
+
+        const testTargetPathExists = await fse.pathExists(testTargetPath);
+        expect(testTargetPathExists).toBeTruthy();
+
+        const testDbmsInfo = (await testEnv.dbmss.info([dbmsId])).toArray();
+        expect(testDbmsInfo[0].id).toEqual(dbmsId);
```
nglgzz, line 170:
```suggestion
        expect(testDbmsInfo[0].id).toEqual(dbmsId);
        expect(testDbmsInfo[0].isCustomInstallation).toBe(true);
```

packages/common/src/entities/dbmss/install.test.ts
```diff
@@ -129,6 +130,71 @@ describe('LocalDbmss - install', () => {
...
+                false,
+                false,
+                testTargetPath,
+            ),
+        ).rejects.toThrowError(
+            new TargetExistsError(`Unable to install to non-empty target directory: ${testTargetPath}`),
+        );
+
+        const testTargetPathExists = await fse.pathExists(testTargetPath);
+        expect(testTargetPathExists).toBeTruthy();
```
nglgzz, line 195: Might be good to also check that the test file is still there.

nglgzz:
Would be also good to have a test case to check that when uninstalling we're deleting the target path as well as the symlink.

packages/common/src/entities/dbmss/install.test.ts
jk05: will do

packages/common/src/entities/dbmss/install.test.ts
jk05: good point

nglgzz (approved):

PR #342, opened 2021-11-22
Use NestMiddleware to register token handlers

nglgzz:

Bug fix

Token handlers are registered during the application initialization (they are express middleware functions), but there is no guarantee that they are registered before the routes. This can cause the routes that are initialized before the handlers to not require authentication when they should. I haven't noticed this happening when running `@relate/web` on its own, but I noticed it happening in certain cases when extending the WebModule. 

### What is the new behavior?
Token handlers are registered using Nestjs middleware, ensuring that they are always registered before any route.

No.

### Other information:
Removes auth for the GraphQL Playground and Swagger API docs. Auth is still needed to be able to fetch types for the autocompletion and docs, but at least it's now possible to load the main interface and pass the token and client ID as HTTP headers from the Playground's UI.

### DIFF
--- a/packages/web/src/auth/auth.module.ts
+++ b/packages/web/src/auth/auth.module.ts
@@ -1,19 +1,40 @@
 import {HttpAdapterHost} from '@nestjs/core';
-import {Inject, Module, OnModuleInit} from '@nestjs/common';
+import {Inject, MiddlewareConsumer, Module, NestModule, OnModuleInit, RequestMethod} from '@nestjs/common';
 import {SystemModule} from '@relate/common';
 
 import {AuthService} from './services/auth.service';
+import {ApiTokenMiddleware} from './middleware/api-token.middleware';
+import {AuthTokenMiddleware} from './middleware/auth-token.middleware';
+import cookieParser from 'cookie-parser';
 
 @Module({
     exports: [AuthService],
     imports: [SystemModule],
     providers: [AuthService],
 })
-export class AuthModule implements OnModuleInit {
+export class AuthModule implements OnModuleInit, NestModule {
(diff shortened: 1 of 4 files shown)


packages/web/src/auth/auth.module.ts
```diff
@@ -1,20 +1,41 @@
...
+            .exclude(
+                {
+                    path: '/graphql',
+                    method: RequestMethod.GET,
+                },
+                {
+                    path: '/api-docs',
+                    method: RequestMethod.GET,
+                },
+            )
```
jk05, line 35: nice

packages/web/src/auth/middleware/api-token.middleware.ts
```diff
@@ -0,0 +1,47 @@
...
+        const origin = req.get('origin');
+        // Use the client URL otherwise fallback to the Relate server URL.
+        // Requests coming from files might contain either 'null' or 'file://' in the Origin header.
+        const requestUrl = origin && origin !== 'null' && new URL(origin).host ? origin : environment.httpOrigin;
+        const requestHost = new URL(requestUrl).host;
+
+        try {
+            await environment.verifyAPIToken(requestHost, clientId, apiToken);
+            next();
+        } catch (e) {
```
jk05, line 41: do you need the `e` here? similarly in the other file?

jk05 (approved):

packages/web/src/auth/middleware/api-token.middleware.ts
nglgzz: It's probably not needed, I just moved this code without changing it.

PR #343, opened 2021-11-30
Check API token on subscriptions 

nglgzz:

Bug fix

API tokens are not checked on GraphQL subscriptions in @relate/web. This is because the express middleware we use to check for tokens doesn't apply to subscriptions. 

### What is the new behavior?
API tokens are now checked when subscriptions are connecting to the Apollo server.

We're now using `graphql-ws` for subscriptions. In development we're also enabling `subscriptions-transport-ws` to be able to still use the GraphQL Playground or Apollo Explorer (once they add support for `graphql-ws` we can drop the second transport library).

Yes: 

- The `watchInfoDbmss` subscription is now called `watchDbmss`. 
- Clients using subscriptions from @relate/web will have to switch to the `graphql-ws` trasport library.

### Other information:
~~The way we're checking for the API tokens on subscriptions is the same way recommended in the [NestJs documentation](https://docs.nestjs.com/graphql/subscriptions#authentication-over-websocket), but it looks like this check can be bypassed anyway (https://github.com/apollographql/subscriptions-transport-ws/issues/349). Switching to the `graphql-ws` library for subscriptions should fix this issue, but I couldn't get it to work. Will need to look more into this.~~


jk05:
In relation to what you mentioned about `graphql-ws`, wanted to highlight there may be [compatibility issues with GraphQL Playground](https://stackoverflow.com/questions/69178586/nestjs-graphql-subscriptions-not-working-with-graphql-ws) and I made an initial quick attempt but couldn't get it to work either (https://github.com/neo4j-devtools/relate/blob/master/packages/web/src/web.module.ts#L38)

packages/common/src/constants.ts
```diff
@@ -149,7 +149,7 @@ export enum PUBLIC_GRAPHQL_METHODS {
     CREATE_ACCESS_TOKEN = 'createAccessToken',
     GET_DBMS = 'getDbms',
     INFO_DBMSS = 'infoDbmss',
-    WATCH_INFO_DBMSS = 'watchInfoDbmss',
+    WATCH_DBMSS = 'watchDbmss',
```
jk05, line 152: 👍🏽  better name

packages/web/src/web.module.ts
```diff
@@ -37,16 +47,48 @@ export interface IWebModuleConfig extends ISystemModuleConfig {
...
+                            const apiToken = Str.from(connectionParams[API_TOKEN_HEADER]).toString();
+                            const environment = await systemProvider.getEnvironment();
+
+                            if (!environment.requiresAPIToken) {
+                                return true;
+                            }
+
+                            // @ts-ignore ws contains the upgradeReq attribute,
+                            // but the type definition for it doesn't.
+                            const origin = ws.upgradeReq.headers?.origin || 'null';
```
jk05, line 71: `'null'` as a string?

packages/web/src/web.module.ts
```diff
@@ -35,18 +43,27 @@ export interface IWebModuleConfig extends ISystemModuleConfig {
...
+                subscriptions: {
+                    // @todo - This is temporarily enabled in development
+                    // because graphql playground and apollo explorer don't yet
+                    // support `graphql-ws`. It shouldn't be enabled in
+                    // production because `subscriptions-transport-ws` has a bug
+                    // that will allow certain connections to skip the
+                    // onConnection handler, bypassing the token check.
+                    'subscriptions-transport-ws':
+                        process.env.NODE_ENV === 'development' ? subscriptionsTransportWs(systemProvider) : false,
+                    'graphql-ws': graphqlWs(systemProvider),
```
jk05, line 62: Im not sure how it works together - there wont be a clash in `development` environment here?

nglgzz: According to the Nest.js docs both can be enabled at the same time: https://docs.nestjs.com/graphql/subscriptions#enable-subscriptions

I tried it and it seems to work. I'm guessing that since they use different protocols, the right protocol is used depending on what the client is using.

packages/web/src/web.module.ts
nglgzz: Yeah it's just to simplify the check on the next line, as the origin header can also return null as a string.

jk05 (approved):

PR #344, opened 2021-12-13
Log4j 0-day exploit fix

linuslundahl:
Adds `dbms.jvm.additional=-Dlog4j2.formatMsgNoLookups=true` to the config to protect users from the Log4j 0-day exploit.

Log4j has a 0-day exploit which is present in Neo4j 4.2.x and up. To help protect users we automatically add the setting `dbms.jvm.additional=-Dlog4j2.formatMsgNoLookups=true` to the users `neo4j.conf` before starting the DBMS.

x

### What is the new behavior?
x

x

### Other information:

Quote from Jonatan Jäderberg regarding adding this to every DBMS:

"It is conceivable that some enterprise customer somewhere is actually using this as a feature--i.e., they have some LDAP server with Java code that they want to be involved in their logging.

If that's the case, their particular use case could break.

But in the interest of helping the vast vast majority of users stay safe, I think that's worth it

They can easily override this behavior"

### DIFF
--- a/packages/common/src/entities/dbmss/dbmss.local.ts
+++ b/packages/common/src/entities/dbmss/dbmss.local.ts
@@ -298,6 +298,17 @@ export class LocalDbmss extends DbmssAbstract<LocalEnvironment> {
         return List.from(nameOrIds)
             .mapEach(async (nameOrId) => {
                 const {id} = await this.getDbms(nameOrId);
+                const {version} = await this.get(nameOrId);
+
+                // Add settings that protects users from Log4j 0-day exploit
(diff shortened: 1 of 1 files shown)


nglgzz (approved):
:sparkles:

oskarhane:
This might not work for Neo4j 4.2.X. See thread: https://neo4j.slack.com/archives/C02R45A8308/p1639401028109500

linuslundahl:
> This might not work for Neo4j 4.2.X. See thread: https://neo4j.slack.com/archives/C02R45A8308/p1639401028109500

Seems to be under specific conditions...

PR #345, opened 2022-02-01
Allow URLs for Sample Project dumps and scripts

linuslundahl:

Allows for `dumpFile` or a `scriptFile` in the sample project dbms description to point to a URL instead of a local file. The file will be downloaded before it is imported.

Also removes `got` for `listSampleProjects` and uses `requestJson` instead, which should work with the Desktop proxy. 

Currently only local files can be defined.

### What is the new behavior?
(if this is a feature change)

### DIFF
--- a/packages/common/src/entities/projects/projects.abstract.ts
+++ b/packages/common/src/entities/projects/projects.abstract.ts
@@ -1,5 +1,4 @@
 import {List} from '@relate/types';
-import {Got} from 'got';
 
 import {EnvironmentAbstract} from '../environments';
 import {ManifestAbstract} from '../manifest';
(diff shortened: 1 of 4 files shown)


packages/common/src/entities/projects/projects.abstract.ts
```diff
@@ -130,7 +129,7 @@ export abstract class ProjectsAbstract<Env extends EnvironmentAbstract> {
     /**
      * Lists sample projects from github (https://github.com/neo4j-graph-examples)
      */
-    abstract listSampleProjects(fetch?: () => any | Got): Promise<List<ISampleProjectRest>>;
```
nglgzz, line 133: :+1:

nglgzz (approved):
Looks good to me :rocket:

PR #348, opened 2022-03-08
Don't add log4j vulnerability fix every time.

linuslundahl:

Prevent duplication of the log4j vulnerability fix setting. Apparently doing `neo4jConfig.get('dbms.jvm.additional=-Dlog4j2.formatMsgNoLookups')` doesn't return `true`, I'm guessing the double equal signs in the setting (`dbms.jvm.additional=-Dlog4j2.formatMsgNoLookups=true`) screws up the fetching of it.

`dbms.jvm.additional=-Dlog4j2.formatMsgNoLookups=true` is added as a new line in the config on each DBMS start.

### What is the new behavior?
`dbms.jvm.additional=-Dlog4j2.formatMsgNoLookups=true` is only added once to the config.


packages/common/src/entities/dbmss/dbmss.local.ts
```diff
@@ -306,7 +306,12 @@ export class LocalDbmss extends DbmssAbstract<LocalEnvironment> {
                         path.join(this.getDbmsRootPath(id), NEO4J_CONF_DIR, NEO4J_CONF_FILE),
                     );
 
-                    neo4jConfig.set('dbms.jvm.additional=-Dlog4j2.formatMsgNoLookups', 'true');
+                    const log4jVulnFix = neo4jConfig.config.find(
+                        (item) => item[0] === 'dbms.jvm.additional' && item[1] === '-Dlog4j2.formatMsgNoLookups=true',
+                    );
```
linuslundahl, line 311: `neo4jConfig.get('dbms.jvm.additional=-Dlog4j2.formatMsgNoLookups')` isn't working when trying to fetch if the setting has been added.

nglgzz: I realized only now, but wouldn't the key be only `dbms.jvm.additional`? So you'd check with 

```
neo4jConfig.get('dbms.jvm.additional') === '-Dlog4j2.formatMsgNoLookups'
```

and set as 

```
neo4jConfig.set('dbms.jvm.additional', '-Dlog4j2.formatMsgNoLookups')
```

linuslundahl: No, there are multiple entries that starts with `dbms.jvm.additional` in the conf file already. And using `neo4jConfig.get('dbms.jvm.additional')` will return the first one. In the default case we get `-XX:+UseG1GC`. So that won't do...

linuslundahl: and the actual setting is: `dbms.jvm.additional=-Dlog4j2.formatMsgNoLookups=true`.

nglgzz (approved):

PR #359, opened 2022-06-23
Delete remote environment

nglgzz:

Delete a feature.

Currently two types of environment can be created:

- Local: this is an environment where Relate has access to the entities' files and can manage them directly. 
- Remote: this is an environment that communicates to another instance or Relate through the GraphQL server in @relate/web, and manages the entities through the GraphQL API.

While this feature is technically awesome, it's incomplete, it's not being used, it's very unlikely that it will ever be used, and it adds some complexity to the codebase. So I'm deleting it to get rid of the extra complexity and dependencies.

### What is the new behavior?
Remote environments are no longer supported. Only local environments can be managed with Relate.

Yes. Remote environments are no longer supported.

### Other information:

It might be good to get rid of the environment types altogether to further simplify the codebase, but I'll leave this for a future PR.


linuslundahl (approved):
Good initiative! LGTM!

PR #386, opened 2022-08-15
Remove electron package

nglgzz:

Remove `@relate/electron`. This package is currently a wrapper of `@relate/web` that doesn't add extra functionality but complicates testing and maintenance, so we're getting rid of it. If there is a need for it in the future we can bring it back.

Yes. The `@relate/electron` package is now removed and needs to be replaced with `@relate/web`.


linuslundahl (approved):
🗑️ LGTM

PR #419, opened 2022-09-22
Adjust list DBs result.

linuslundahl:
The result from the SHOW DATABASES command has changed and this addresses that.

The result for `SHOW DATABASES` has changed in 5.*, this PR adds support for that.

Currently the result only supports 4.*.

### What is the new behavior?
Support for both 4.* and 5.*.

-

### DIFF
--- a/packages/common/src/entities/dbs/dbs.local.ts
+++ b/packages/common/src/entities/dbs/dbs.local.ts
@@ -104,22 +104,8 @@ export class LocalDbs extends DbsAbstract<LocalEnvironment> {
         javaPath?: string,
     ): Promise<string> {
         const dbms = await this.environment.dbmss.get(nameOrId);
-
-        this.environment.dbs.exec(nameOrId, `STOP DATABASE ${database}`, {
-            user: 'neo4j',
-            database: 'neo4j',
-            accessToken: credentials,
-        });
-
         const params = ['dump', `--database=${database}`, `--to=${to}`];
         const result = await neo4jAdminCmd(await this.resolveDbmsRootPath(dbms.id), params, '', javaPath);
-
-        this.environment.dbs.exec(nameOrId, `START DATABASE ${database}`, {
-            user: 'neo4j',
-            database: 'neo4j',
-            accessToken: credentials,
-        });
-
         return result;
     }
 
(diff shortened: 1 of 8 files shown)


packages/common/src/entities/dbs/dbs.local.ts
```diff
@@ -59,21 +62,64 @@ export class LocalDbs extends DbsAbstract<LocalEnvironment> {
...
+    async dump(
+        nameOrId: string,
+        database: string,
+        to: string,
+        credentials: string,
+        javaPath?: string,
+    ): Promise<string> {
         const dbms = await this.environment.dbmss.get(nameOrId);
+
+        this.environment.dbs.exec(nameOrId, `STOP DATABASE ${database}`, {
```
nglgzz, line 108: I'm afraid that changing the state of the DB will make creating dumps less reliable since if there is an error, the DB will be left offline. And one could go on to do other things that might make it harder to successfully start the DB again (eg. running an upgrade).

Also, this would only work when the DBMS is started and an accessToken was created for the `neo4j` user.

packages/common/src/entities/dbs/dbs.local.ts
```diff
@@ -59,21 +62,64 @@ export class LocalDbs extends DbsAbstract<LocalEnvironment> {
         );
 
         return dbs.mapEach((db) => {
+            if (
+                dbms.version &&
+                semver.satisfies(dbms.version, NEO4J_JAVA_17_VERSION_RANGE, {includePrerelease: true})
```
nglgzz, line 67: Might be good to create a NEO4J_VERSION_5X constant for this

linuslundahl: Should we rename `NEO4J_JAVA_17_VERSION_RANGE` to `NEO4J_VERSION_5X`?

packages/common/src/entities/dbs/dbs.local.ts
linuslundahl: Oh, good that you caught that, this accidentally committed.

packages/common/src/entities/dbs/dbs.local.ts
nglgzz: Sounds good :+1:

nglgzz (approved):

PR #420, opened 2022-09-23
Gitignore .npmrc file

nglgzz:
### Please check if the PR fulfills these requirements
- [x] The commit message follows our [guidelines](https://github.com/neo4j-devtools/relate/blob/master/CONTRIBUTING.md)
- [ ] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)

### What kind of change does this PR introduce?
Gitignore the .npmrc file that is modified during the release process, and remove the git hook for running e2e tests on push.

### What is the current behavior?
The release pipeline is failing to create the publish commit because the working directory is unclean after we update the npmrc file.

### Does this PR introduce a breaking change?
No

PR #421, opened 2022-09-23
Update global .npmrc instead of local one on release action

nglgzz:

Git can't ignore a tracked file, so the change from #420 is not working. This results in the release pipeline still failing because the working is unclean after we update the .npmrc file. 

This PR reverts ignoring the .npmrc file (which didn't work) and updates the release action to modify the user npm config instead of the local one. 

No


linuslundahl (approved):
👍

PR #441, opened 2022-11-23
Tweak watchDbmss to observe changes in the server status rather then the process status

nglgzz:
- [ ] The commit message follows our [guidelines](https://github.com/neo4j-devtools/relate/blob/master/CONTRIBUTING.md)

- `infoDbmss` no longer requires an array of IDs, and when no ID is provided we return the entire list of DBMSs and their status. This is to avoid having to send two separate requests for getting the status of all DBMSs.

- `watchDbmss` now sends events for STARTED and STOPPED **after** we confirmed that the Neo4j server status is online or offline. The previous behavior would react to the change in the file system and send whatever status we had at the time the change happened. That could result for example in a STARTED event being created after the DBMS process started (the PID file was created) but before the server was online.

No

### DIFF
--- a/.changeset/large-poets-fail.md
+++ b/.changeset/large-poets-fail.md
@@ -1,0 +1,5 @@
+---
+'@relate/web': patch
+---
+
+Tweak watchDbmss subscription to observe server rather than process status. No longer require IDs array on infoDbmss query.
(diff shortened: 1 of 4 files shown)


linuslundahl (approved):
LGTM

PR #449, opened 2023-01-05
Neo4j 5.x upgrade fixes.

linuslundahl:

* Bump supported version to 4.1
* Prevent Neo4j 4 -> 5 upgrades.
* Don't run migration step for 5.0 upgrades.

### What is the new behavior?
(if this is a feature change)

### DIFF
--- a/.changeset/cool-vans-change.md
+++ b/.changeset/cool-vans-change.md
@@ -1,0 +1,5 @@
+---
+'@relate/common': patch
+---
+
+Add support for upgrades between 5.x versions.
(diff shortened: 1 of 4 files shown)


packages/common/src/utils/system/test-environment.ts
```diff
@@ -19,11 +19,11 @@ export const TEST_NEO4J_CREDENTIALS = 'password';
 export const TEST_NEO4J_VERSIONS = {
     default: process.env.TEST_NEO4J_VERSION ?? '4.4.11',
 
-    majorUpgradeSource: '3.5.34',
-    majorUpgradeTarget: '4.0.12',
+    majorUpgradeSource: '4.0.12',
```
nglgzz, line 22: This version is used for testing major upgrades from 3 to 4. If we're dropping support for version 3, I think we should remove this upgrade path as well and the tests for it.

linuslundahl: Ah, right... Good point! I think we still need to allow upgrading from 3 to 4...

nglgzz (dismissed):
Looks good to me

nglgzz:
Ah forgot, but you need to add a changeset if you want to release this.

nglgzz (approved):

